PR #39, opened 2017-10-23
Add LICENSE

izqui:


izqui:
Would never have expected adding the license would have that effect on coverage 😂

onbjerg:
Very nice, thanks for the additional coverage 💃

PR #44, opened 2018-01-08
Refactor APM component to use `ethereum-ens`

### LINKED ISSUE #42: Use `ethereum-ens`
There's no reason we should manually maintain ENS-specific code if there exists a good and minimal library for it.

onbjerg:
Closes #42


onbjerg:
Not needed anymore

PR #50, opened 2018-02-16
Notify: add date parameter

sohkai:
Adds an optional date parameter, as a lot of notifications will only be read after the fact (assuming you won't have the dapp open all the time).


onbjerg:
I think `context` should be first, since that will probably(?) be used most frequently

sohkai:
I just assumed `context` would always be the last parameter, but sounds good.

PR #53, opened 2018-02-22
Add MessagePort message provider

sohkai:
Renames the original `PostMessage` provider to `WindowMessage` (since `postMessage()` across windows is the only one that I know of using the `targetOrigin` second parameter).

Note that using `postMessage()` across web workers doesn't seem to send an event with an `event.source`, even though [MDN seems to think it should](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/source). This is in both Chrome and Firefox.

### DIFF
--- a/packages/aragon-messenger/src/providers/WindowMessage.js
+++ b/packages/aragon-messenger/src/providers/WindowMessage.js
@@ -2,13 +2,13 @@ import Provider from './Provider'
 import { Observable } from 'rxjs/Rx'
 
 /**
- * A provider that uses the PostMessage API to pass messages between frames and WebWorkers.
+ * A provider that uses the Window postMessage API to pass messages between windows.
  *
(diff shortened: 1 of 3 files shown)


packages/aragon-messenger/src/providers/WindowMessage.js
```diff
@@ -17,12 +17,12 @@ export default class PostMessage extends Provider {
...
    * An observable of messages being sent to this provider.
    *
-   * @memberof PostMessage
+   * @memberof WindowMessage
    * @instance
    * @returns {Observable}
    */
   messages () {
-    return Observable.fromEvent(window, 'message')
+    return Observable.fromEvent(window, 'message', false)
```
sohkai: The `false` bit passes `false` into `addEventListener()`'s last argument. It's not strictly necessary for newer browsers, but might as well include it.

onbjerg:
👌

PR #59, opened 2018-02-28
Implement beta templates

### LINKED ISSUE #22: Add support for DAO factories

izqui:
Requires #72

See [`packages/aragon-wrapper/sandbox.js`](https://github.com/aragon/aragon.js/pull/59/files#diff-dbc261c18771de6d8ab735e0c15a8e63) for how to use them.

Requires https://github.com/aragon/aragon-apps/pull/115 and local linking to `@aragon/templates-beta`

Closes #22


onbjerg:
Can we generalize the API a bit more so we have a format to add templates in an array or something, and two or three functions: one to list templates, one to view a specific template (i.e. get the data about the params, title etc.) and one to deploy using a template ID and the filled out params?

izqui:
@onbjerg did a fair amount of changes to the file structure thinking about new templates, let me know what you think. 

Also be careful with merging as it comments out the cache because it would break as there is a lot of code commented for non-browser support. I also had to add `babel-polyfill` to use awaits

packages/aragon-wrapper/src/index.js
```diff
@@ -54,15 +57,16 @@ export default class Aragon {
...
     }))
 
+    this.templates = Templates(this.web3, this.apm, options.from)
+
     // Set up the kernel proxy
     this.kernelProxy = makeProxy(daoAddress, 'Kernel', this.web3)
 
     // Set up cache
-    this.cache = new Cache(daoAddress)
+    // this.cache = new Cache(daoAddress)
```
onbjerg: Is this supposed to be committed?

packages/aragon-wrapper/src/sandbox.js
```diff
@@ -1,28 +1,24 @@
+import 'babel-polyfill'
```
onbjerg: Can you delete this file?

packages/aragon-wrapper/src/templates/index.js
```diff
@@ -0,0 +1,78 @@
...
+const sleep = t => new Promise(r => setTimeout(() => r(), 1000 * t))
+
+module.exports = (web3, apm, from) => {
+  const newToken = async (template, name) => {
+    const receipt = await template.methods.newToken(name, name).send({ from, gas: 4e6 })
+    return receipt.events.DeployToken.returnValues
+  }
+
+  const newInstance = async (template, name, params) => {
+    await sleep(2) // ensure newToken is submitted before
```
onbjerg: Can you chain the promises instead of doing this?

packages/aragon-wrapper/src/index.js
onbjerg: Never mind, saw your comment 👍

onbjerg:
Will merge this after I fix the comments I put here and also I will fix the cache first (should be done soon-ish)

onbjerg:
Requires #72

bpierre:
@onbjerg I’m still working on this, I will add some commits today

onbjerg:
Lemme know when this is ready for review again @bpierre

bpierre:
@onbjerg Ready!

PR #61, opened 2018-03-04
Notifications: acknowledge, clear, and add sorted by datetime

sohkai:
Requires https://github.com/aragon/aragon.js/pull/60.

Adds ability to acknowledge and clear notifications (naive inefficient implementation).


packages/aragon-wrapper/src/index.js
```diff
@@ -197,31 +199,114 @@ export default class Aragon {
...
+        if (notification && !notification.read) {
+          notification.acknowledge = () => this.acknowledgeNotification(notification.id)
+        }
+      })
+    }
+
+    this.notifications = new BehaviorSubject(cached)
+      .scan((notifications, { modifier, notification }) => modifier(notifications, notification))
+      .do((notifications) => this.cache.set('notifications', notifications))
+      .publishReplay(1)
```
sohkai: Adding `.publishReplay()` seems to turn the `BehaviourSubject` into a multicasted observable, such that all observers (even those that subscribe later) receive the same values (rather than only the last one sent through the subject). Not 100% sure this is correct, though.

packages/aragon-wrapper/src/index.js
```diff
@@ -197,31 +199,114 @@ export default class Aragon {
...
+        const newNotificationIndex = notifications.findIndex(
+          notification => !isBefore(new Date(notification.date), date)
+        )
+        return newNotificationIndex === -1 ?
+          [...notifications, notification] :
+          [
+            ...notifications.slice(0, newNotificationIndex),
+            notification,
+            ...notifications.slice(newNotificationIndex)
+          ]
```
sohkai: The notifications are inserted sorted by date time; I thought it was better to do this work here on every insert rather than having the frontend re-sort the array on each update.

packages/aragon-wrapper/src/index.js
onbjerg: `publishReplay` wraps the observable in a `ReplaySubject`, but a `BehaviorSubject` is already a `ReplaySubject`, so it's pretty verbose to have it.

http://reactivex.io/rxjs/test-file/spec-js/subjects/BehaviorSubject-spec.js.html#lineNumber106

onbjerg:
Will look at this later today

packages/aragon-wrapper/src/index.js
sohkai: The behaviour of `BehaviourSubject` is weird when a `scan()`'s involved (`scan` only accumulates after you've subscribed):

```js
const subject = new rxjs.BehaviorSubject(0).scan((acc, val) => acc + val, 0)
subject.subscribe(console.log) // prints 0
subject.next(1) // prints 1
subject.next(1) // prints 2
subject.subscribe(console.log) // prints 2, 1
subject.next(1) // prints 3, 2
```

Using `publishReplay()` seems to solve this:

```js
const subject = new rxjs.BehaviorSubject(0).scan((acc, val) => acc + val, 0).publishReplay(1)
subject.connect()
subject.subscribe(console.log) // prints 0
subject.next(1) // prints 1
subject.next(1) // prints 2
subject.subscribe(console.log) // prints 2, 2
subject.next(1) // prints 3, 3
```

onbjerg: Weird! Nice catch :)

PR #65, opened 2018-03-05
Cache: send current value when observing

sohkai:
With `startWith`, `observe()` would only send an initial value the next time the cache is updated (which is surprising, I think).


onbjerg:
I am not sure what the actual change is, `defaultValue` already exists, no?

PR #66, opened 2018-03-06
Use arapp.json instead of artifact.json

sohkai:
I think we changed it some time ago to `arapp.json`.


onbjerg:
No, arapp.json is used to generate artifact.json

PR #67, opened 2018-03-07
Messaging: default response payloads to null

sohkai:
Currently the cache is broken, as the initial payload is always `undefined`.

~~Not sure if there's any specifics for JSON-RPC, like not allowing undefined payloads.~~

[See below](https://github.com/aragon/aragon.js/pull/67#issuecomment-371220811): apparently JSON RPC doesn't allow `undefined`, so next best is to let the response generator default the payload to `null` if it's not given or is given as `undefined`.

### DIFF
--- a/packages/aragon-messenger/src/jsonrpc.js
+++ b/packages/aragon-messenger/src/jsonrpc.js
@@ -9,7 +9,7 @@ export const encodeRequest = (method, params = []) => {
   }
 }
 
-export const encodeResponse = (id, result) => {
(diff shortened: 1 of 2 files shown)


onbjerg:
Ping @sohkai

sohkai:
> JSONRPC does not allow undefined payloads

Ah, thought it might've been something like this. Then yes, we should default to `null` on any payload.

sohkai:
@onbjerg I've repurposed this PR to instead default values to `null` if they're not given / undefined when making a response.

PR #68, opened 2018-03-07
Client: make store observable hot

sohkai:
I got bit by the hot / cold observables flu so I thought it might be less confusing if `store()` was always hot.

Probably biased from past experiences / other relatables, but I feel like a "store" should immediately work (rather than on an explicit subscription).

🤷‍♂️

### DIFF
--- a/packages/aragon-client/src/index.js
+++ b/packages/aragon-client/src/index.js
@@ -109,7 +109,7 @@ class AppProxy {
       .publishReplay(1)
     store$.connect()
 
-    return store()
+    return store$
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/aragon-client/src/index.js
```diff
@@ -100,12 +100,16 @@ class AppProxy {
...
+    const store$ = initialState
       .switchMap((initialState) =>
         this.events()
           .mergeScan(wrappedReducer, initialState, 1)
           .map((state) => this.cache('state', state))
       )
+      .publishReplay(1)
+    store$.connect()
+
+    return store()
```
onbjerg: This a typo?

onbjerg: Ping @sohkai

sohkai: Oh yes!

Issue #69, opened 2018-03-12
Fix cache for non-browser environments


onbjerg:
This is actually less trivial than I anticipated, since Babel hoists all requires, so doing something like

```js
let adapter
if (isBrowser) {
  const BrowserDatabaseAdapter = require('foo')
  adapter = BrowserDatabaseAdapter()
} else {
  const FileSystemAdapter = require('bar')
  adapter = FileSystemAdapter()
}
```

won't work, even though it should in theory 🤦‍♂️ 

Could probably maybe hopefully be fixed with proper tree shaking

onbjerg:
Might be solved with #70

PR #71, opened 2018-03-12
Use Rollup to bundle library

### LINKED ISSUE #70: Use Rollup for bundling
Multiple reasons

- Smaller bundle sizes (because tree shaking)
- Rollup is nice
- Multiple targets (browsers, node, ...)
- Might fix #69 because of tree shaking

onbjerg:
This PR tree shakes our stuff and should also make #69 solvable.

**To do**
- Generate a non-browser bundle, too (`resolve({ browser: false })`)

Closes #70

### DIFF
--- a/packages/aragon-wrapper/.babelrc
+++ b/packages/aragon-wrapper/.babelrc
@@ -12,7 +12,8 @@
             "android >= 4.2"
           ]
         },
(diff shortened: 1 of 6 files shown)


onbjerg:
Doesn't work

PR #72, opened 2018-03-12
Make the cache work in not-a-browser

### LINKED ISSUE #69: Fix cache for non-browser environments

onbjerg:
So, the primary issue with working in both the browser and outside the browser right now is to make sure that we do not require the filesystem specific adapters for lowdb in the browser, as that will break.

Rather than spend a lot of time on fixing this right now (which I have already done), we're just going to use the memory adapter.

This works fine for things like Aragon CLI, but will be a pain when we start using Electron, unless Electron also supports localStorage (which I think it does).

Closes #69


onbjerg:
Erhh, this needs a bit more work. I need to build a non-browser bundle as well (in #71)

onbjerg:
This works now

PR #74, opened 2018-03-13
Add build and build:watch global tasks

bpierre:
Will make it easier to build / watch the three libraries together.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,8 @@
     "lint": "lerna run lint",
     "test": "lerna run test",
     "install": "npm run bootstrap",
-    "bootstrap": "lerna bootstrap"
+    "bootstrap": "lerna bootstrap",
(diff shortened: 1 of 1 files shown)


sohkai:
❤️  x 100

package.json
```diff
@@ -6,6 +6,8 @@
     "lint": "lerna run lint",
     "test": "lerna run test",
     "install": "npm run bootstrap",
-    "bootstrap": "lerna bootstrap"
+    "bootstrap": "lerna bootstrap",
+    "build": "lerna run build",
+    "build:watch": "lerna run --parallel build -- -- --watch"
```
bpierre: Note: the first `--` is to separate lerna args from the npm command args, and the second is to separate the npm args from the script args.

onbjerg:
👏

Issue #75, opened 2018-03-13
Allow calls to arbitrary contracts


sohkai:
Ara-clients can't use web3 directly, so we need to expose a way for them to read state from other contracts.

It would also be nice to have a way to get an event stream from another contract. Even better if this was somehow integrated into the `store()` (e.g. upon initializing the store, you provided an array of other event streams that were then merged with the app contract's events).

onbjerg:
Duplicates #43

Issue #43, opened 2017-11-20
Allow to make calls and send transactions to contract addresses other than the AppProxy


izqui:
Related to #41.

Apps that interact with contracts outside an Aragon organization need the ability to make calls and transactions to other contracts. The most obvious example is apps that work with tokens (such as the Token Manager or Finance) which depend directly on the tokens to fetch balances. Another example is the Fundraising app, in which buyers need to interact with their tokens in order to buy in sales.

In the case of sending transactions, it should be handled differently than current 'intents' and allow to perform the action directly without calculating any fwd path. The signer in the wrapper should let the user know it is not a 'normal action' being performed and it could have consequences. In the case of known actions such as token transfers, we would be able to provide a description for it.

sohkai:
From #75:

> It would also be nice to have a way to get an event stream from another contract. Even better if this was somehow integrated into the store() (e.g. upon initializing the store, you provided an array of other event streams that were then merged with the app contract's events).

sohkai:
Closing for #120 (send-support), as #76 has implemented read-support.

PR #76, opened 2018-03-14
External contracts

### LINKED ISSUE #41: Allow listening to arbitrary events emitted by addresses other than the proxy
Some apps need to get events from other addresses, for example the TokenManager app doesn't emit events on token transfers because the managed token already emits them.

This should also allow to pass a filter to take advantage of indexed event logs.

onbjerg:
This PR allows you to interact with external contracts (i.e. contracts that are assumed to not be an Aragon app), such as tokens.

It allows you to perform calls and filter for events, but provides no automatic caching or convenience functions.

**Usage-ish**

```js
const token = app.external(address, jsonInterface)

// A call
// Returns an Observable that emits a single value (the return value, a Web3 `Result` object)
token.symbol()

// Events
// Returns an Observable that emits Web3-structured events over time
token.events()
```

Note that you need to supply (at least a partial) JSON interface to interact with the contract.

**Example**

As a minimum viable example for getting the symbol of a token (conforming to Human Standard Token) and setting the app identifier to be that symbol, this would suffice:

```js
const Aragon = require('@aragon/client')

const app = new Aragon()
const token = app.external('0xdeadbeef', [
  {
    'constant': true,
    'inputs': [],
    'name': 'symbol',
    'outputs': [
      {
        'name': 'symbol',
        'type': 'string'
      }
    ],
    'payable': false,
    'type': 'function'
  }
])

token.symbol().subscribe(
  ({ symbol }) => app.identify(symbol)
)
```

**Improvements**

Some future improvements could include

- Transactions to external contracts, possibly with "universal transaction pathing"
- Convenience functions, such as an integration with `store` (see brief explanation in #75)

Closes #41 and partially adresses #43


onbjerg:
Tested locally and working, need a go from either one of you @bpierre @sohkai

packages/aragon-client/src/index.js
```diff
@@ -36,6 +36,45 @@ class AppProxy {
...
 
+  /**
+   * Create a handle to an external contract.
+   *
+   * @param  {string} address The address of the external contract
+   * @param  {Array<Object>} jsonInterface The JSON interface of the external contract
+   * @return {Object}
+   */
+  external (address, jsonInterface) {
+    let contract = {
```
sohkai: Could be `const`?

packages/aragon-client/src/index.js
```diff
@@ -36,6 +36,45 @@ class AppProxy {
...
+      events: (fromBlock = 0) => {
+        return this.rpc.sendAndObserveResponses(
+          'external_events',
+          [
+            address,
+            jsonInterface.filter(
+              (item) => item.type === 'event'
```
sohkai: Cool! I had no idea you could put in partial ABIs and everything would just work.

packages/aragon-client/src/index.js
```diff
@@ -36,6 +36,45 @@ class AppProxy {
...
+          ]
+        ).pluck('result')
+      }
+    }
+
+    // Bind calls
+    const callMethods = jsonInterface.filter(
+      (item) => item.type === 'function' && item.constant
+    )
+    for (let methodABI of callMethods) {
```
sohkai: I do find it slightly odd this isn't a `forEach` or `reduce`

packages/aragon-client/src/index.js
onbjerg: yeah that's something I missed while iterating a bit

packages/aragon-client/src/index.js
onbjerg: could be a `forEach`, probably also a reduce but would feel a bit weird with reduce imo

PR #77, opened 2018-03-14
Identifiers

onbjerg:
This PR allows apps to feel unique 😭 

**Usage**

```js
app.identify(identifier) // => nothing
```

Wrappers can access apps' identifiers (might be `undefined`!) on the `identifier` key of the apps array emitted by the `wrapper.apps` observable.


packages/aragon-wrapper/src/index.js
```diff
@@ -175,10 +183,39 @@ export default class Aragon {
           ))
         )
       )
+    this.apps = this.appsWithoutIdentifiers
+      .withLatestFrom(
```
sohkai: One thing to note here (it might not work as expected): emitting events from `this.identifiers` won't trigger an update on `this.apps` so a new app would have to be emitted before any changes from new identifiers takes place.

Maybe something like `combineLatest` would be better?

onbjerg: Emitting events from `this.identifiers` will indeed trigger an update, since it is the source observable. Look at the marble diagram [here]().

However, emitting from the `apps` observable will not. This is an overlook, and `combineLatest` would solve it.

onbjerg:
@sohkai Fixed in 6e4b595. Tested this locally in sandbox and it does attach identifiers 🎉

packages/aragon-wrapper/src/index.js
```diff
@@ -184,13 +178,18 @@ export default class Aragon {
...
-      .withLatestFrom(
-        this.identifiers,
+      .combineLatest(
+        this.identifiers.scan(
+          (identifiers, { address, identifier }) =>
+            Object.assign(identifiers, { [address]: identifier }),
+          {}
+        ).startWith({}),
         function attachIdentifiers (apps, identifiers) {
+          console.log(apps, identifiers)
```
sohkai: Probably forgot to remove this :)

onbjerg: whoops

PR #78, opened 2018-03-14
Allow client store to take in other event streams

sohkai:
Defers `event` observables so they're cold until subscribed to (to avoid losing events), and merges an array of given event observables with the app's events.


onbjerg:
Can you change the target branch to master?

sohkai:
Replaced by #92.

PR #82, opened 2018-03-14
Client: re-export providers from messenger

sohkai:
In the app UIs, we still need the providers to get the `WindowProvider`, so we might as well re-export it :).


onbjerg:
fastest merge ever

PR #84, opened 2018-03-14
Client: change proxy interface to take in a provider instead of a messenger

sohkai:
Makes this mirror the wrapper's `runApp()`, and allows us to not have to pull in the `Messenger` in clients if we need to change the provider.


onbjerg:
I thought about dong this too, but I was hesistant since I saw that it would break some of the apps right now - are you sure you want to merge this now? If yes, I'll merge it, because it's a good idea.

onbjerg:
Talked about internally, merging now

PR #85, opened 2018-03-15
Block cache

### LINKED ISSUE #63: Cache transaction index in addition to block number
For event processing caching only the block number is not enough as the client can be disconnected in the middle of processing a given block. 

In the context of block reorg, we should maybe also cache all the processed txids in a given block and revert those if we detect the ordering ever changing, but this is a whole other can of worms

onbjerg:
Introduces the concept of cursors.

A cursor is comprised of a block number, a transaction index in that block, and a log index in that block that we have currently processed.

If no cursor has been saved, we default to start from the initialization block of the proxy.

If the proxy does not have a `getInitializationBlock` method, we default to start from block number 0, but this should probably be set to some sane default. A sane default is hard to find, since testnets might be restarted, starting from block 0 again. A solution could be to either assume that does not happen, or to have a sane default block per network.

Every time an event has been handled in the client, it will also save the current cursor. This cursor is saved in cache and is thus readable.

**A thing to note** is that cursors are only saved automatically if you use `app#store`.

Closes #63 and #6

### DIFF
--- a/packages/aragon-client/src/index.js
+++ b/packages/aragon-client/src/index.js
@@ -146,6 +146,18 @@ class AppProxy {
   store (reducer) {
     const initialState = this.state().first()
 
+    const cacheStateAtCursor = ({ blockNumber, transactionIndex, logIndex }) =>
+      (state) => {
+        // Cache cursor
+        this.cache('cursor', {
+          blockNumber,
+          transactionIndex,
+          logIndex
+        })
+
+        return state
+      }
+
     // Wrap the reducer in another reducer that
     // allows us to execute code asynchronously
     // in our reducer. That's a lot of reducing.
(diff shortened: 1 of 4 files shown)


chris-remus:
Nice to have at the moment.

Quazia:
Awesome great to hear. I think this just needs review but it's probably going to require some effort to update/merge. Does that mean this is de-prioritized until after Devcon?

sohkai:
@Quazia Yes, the main concern is with how this will handle #38. Caching could be quite dangerous in the case of a re-org and you have no way to roll back or detect new state.

sohkai:
This has been superseded by #297, which uses a different mechanism to cursors (more of a transactional "commit" strategy).

PR #93, opened 2018-03-16
wrapper: Allow wrapper to set accounts manually

### LINKED ISSUE #62: Add a way to get the addresses the user controls
This might be good for *directed notifications*, i.e. notifications that are only pertinent to a single user. Other use cases probably.

### LINKED ISSUE #91: Add ability to set available accounts
We're currently using two different instances of web3 (metamask provider for accounts, another for aragon.js events), so aragon.js doesn't know the real accounts of the user.

It'd be nice if we could manually set this on the wrapper; if it hasn't been set, it should default to its own `web3.eth.accounts`.

Breaks transaction pathing on the client :(.

onbjerg:
Closes #91 

Also adds `accounts` RPC message, closes #62


packages/aragon-wrapper/src/index.js
```diff
@@ -419,12 +419,26 @@ export default class Aragon {
     }
   }
 
+  /**
+   * Set the available accounts for the current user.
+   *
+   * @param {Array<string>} accounts
+   * @return {void}
+   */
+  setAccounts (accounts) {
```
sohkai: A little bit confused about the type `accounts` should be; `getAccounts()` apparently returns a `Promise` but `setAccounts()` takes in an array?

onbjerg: `getAccounts` returns a promise because it falls back to calling via Web3, which returns a promise.

I will move `accounts` as an optional parameter to `init`, but `getAccounts` will always be a promise.

onbjerg:
works w tx pathing in sandbox

PR #95, opened 2018-03-18
wrapper: RxJS-ify templates

onbjerg:
Returns an observable from the `newDAO` function of the templates.

This observable will emit events of this shape:

```js
{
  transaction: 'TOKEN' | 'DAO'
  status: 'SIGNED' | 'MINED' | 'ERROR',
  meta: {
    transactionHash: string,
    ?message: string,
    ?address: string
  }
}
```

Message is only set on `ERROR`, address is only set on `MINED`.

The observable will eventually complete when
- The transactions error out, or
- Both transactions are mined

The observable will also emit an error if the contract address of the template was not found, so it is essential to subscribe like this for completeness:

```js
templates.newDAO().susbcribe(
  console.log, console.error, console.log
)
```


packages/aragon-wrapper/src/templates/index.js
```diff
@@ -26,49 +27,118 @@ const templates = {
...
+    })
+
+    return progress
   }
 
   const newInstance = async (template, name, params) => {
-    const receipt = await template.methods.newInstance(name, ...params).send({ from, gas: 6.6e6, gasPrice: minGasPrice })
-    return receipt.events.DeployInstance.returnValues
+    const progress = new Subject()
+    const events = template.methods.newInstance(name, ...params).send({ from, gas: 6.9e6, gasPrice: minGasPrice })
```
sohkai: @izqui Don't remember the exact setting we wanted for gas limit, but I remember metamask was flaking at 6.9.

packages/aragon-wrapper/src/templates/index.js
```diff
@@ -26,49 +27,118 @@ const templates = {
...
 
   return {
     newDAO: async (templateName, organizationName, params) => {
       const tmplObj = templates[templateName]
 
-      if (!tmplObj) throw new Error("No template found for that name")
-
-      const contractAddress = await apm.getLatestVersionContract(tmplObj.appId)
+      if (!tmplObj) {
+        throw new Error('No template found for that name')
```
sohkai: It feels weird that this would throw a normal error, but the next error would be captured in the observable.

It might be more intuitive for them both to have the same behaviour.

packages/aragon-wrapper/src/templates/index.js
onbjerg: When I resolve the conflict it's going to use gas estimation

packages/aragon-wrapper/src/templates/index.js
onbjerg: Mhh, the other errors are transaction specified, so in my mind it would be weird that you could also get an error for the parameters you input

chris-remus:
Nice to have.

sohkai:
To be reworked as part of the new 0.8 onboarding.

PR #104, opened 2018-04-24
Update to ERC897 proxy interface

### LINKED ISSUE #103: Update for ERC897 proxies
`getCode()` -> `implementation()` in the abi and wrapper.

sohkai:
Fixes #103.

I've just changed the two usages of `getCode()` since I'm not sure if the goal of the `abi/`s is to be autogenerated by a tool or eventually exported by aragonOS. In any case, it should be noted that the abi of `AppProxy` is slightly out of sync with the current implementation in aragonOS (`AppProxyBase`; most notably, `isForwarder()` isn't available).

### DIFF
--- a/packages/aragon-wrapper/abi/aragon/AppProxy.json
+++ b/packages/aragon-wrapper/abi/aragon/AppProxy.json
@@ -142,7 +142,7 @@
   {
     "constant": true,
     "inputs": [],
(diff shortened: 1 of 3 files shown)


onbjerg:
Merged in #106, going to re-review this one and merge + release it

PR #106, opened 2018-05-08
Make ERC897 changes backwards compatible

sohkai:
Realized with #104 that we'd be breaking all currently deployed apps on Rinkeby 😅.

f355766 provides a fallback hatch to the old proxy implementation, so all aragonOS 3.x apps work.
d1862bb provides slightly better error handling, and avoids including the kernel proxy in the list of apps.

In the future, we could version aragon.js in step with aragonOS, but we should absolutely **AVOID** breaking ABI changes in non-major versions.

Long term, a way to handle these ABI changes from different aragonOS versions is would be to create a connector module that's highly-coupled to aragonOS and let frontend users choose which version of aragonOS they want to use (e.g. we might be on aragonOS 4, but all their apps are still on 3).

### DIFF
--- a/packages/aragon-wrapper/abi/aragon/AppProxy.json
+++ b/packages/aragon-wrapper/abi/aragon/AppProxy.json
@@ -142,6 +142,20 @@
   {
     "constant": true,
     "inputs": [],
+    "name": "getCode",
+    "outputs": [
+      {
+        "name": "",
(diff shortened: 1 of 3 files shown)


onbjerg:
Good thoughts. I'll review this and merge it in, then merge #104 and release a new version.

PR #110, opened 2018-05-21
Update web3 to 1.0.0.-beta.34

bingen:
Subscirption to events was not working with 1.0.0.-beta.21. See
discussion here:

https://github.com/aragon/aragon-cli/issues/81


packages/aragon-wrapper/package.json
```diff
@@ -75,7 +75,7 @@
     "radspec": "^1.0.0-beta.7",
     "rxjs": "^5.5.6",
     "uuid": "^3.2.1",
-    "web3": "1.0.0-beta.21",
+    "web3": "^1.0.0-beta.34",
```
sohkai: Actually let's make sure this is pinned.

PR #123, opened 2018-07-12
Migrate to rxjs 6

### LINKED ISSUE #109: Migrate to RxJS 6
See https://auth0.com/blog/whats-new-in-rxjs-6/

0x6431346e:
This fixes #109, but it's still work in progress. Left to do:
- [x] migrate the wrapper src code
- [x] test


sohkai:
Awesome @0x6431346e!!

0x6431346e:
Any thoughts on the testing strategy? :smile: Shall I write some unit tests?

sohkai:
@0x6431346e Testing would be amazing! The main problem we've had with adding tests is mocking an ethereum chain and its events, but definitely there's low hanging fruit for testing a lot of the smaller modules!

packages/aragon-wrapper/src/index.js
```diff
@@ -200,53 +202,56 @@ export default class Aragon {
...
-              return app
+    this.apps = combineLatest(
+      this.appsWithoutIdentifiers,
+      this.identifiers.pipe(scan(
+        (identifiers, { address, identifier }) =>
+          Object.assign(identifiers, { [address]: identifier }),
+        {}
+      )),
+      startWith({}),
+      function attachIdentifiers (apps, identifiers) {
```
sohkai: It looks like the [result selectors are being deprecated with v6](https://rxjs-dev.firebaseapp.com/guide/v6/migration#result-selectors-removed-or-deprecated), what do you think about migrating them over (it looks like it's usually just a pipe to `map` instead)?

0x6431346e: Turns out `apps` is an array in this context

0x6431346e:
This is ready for review :)

sohkai:
@0x6431346e Something I'm worried about with this upgrade is how dependencies (e.g. aragon/aragon and aragon-apps) will be affected. I'm leaning on holding off from merging this until we're in better shape for mainnet and have time to spare for migrating everything over all at once (either before or after the launch).

I think we'll need a few more changes / features in aragon.js to support the mainnet roadmaps for the apps, but hopefully it won't be too much work to update this afterwards.

If the `babel@7` migration has taught me anything, it's that this migration can be really painful :(.

sohkai:
@0x6431346e Wondering what your availability might be like to make a push for this in the next two weeks? I think it'd be really nice to get things frozen in the next few days and then commit to this migration so we can release a new version.

0x6431346e:
> @0x6431346e Wondering what your availability might be like to make a push for this in the next two weeks? I think it'd be really nice to get things frozen in the next few days and then commit to this migration so we can release a new version.

@sohkai I'm available, is it just this PR or do other repos need to be updated as well?

0x6431346e:
@2color do you want me to fix the conflicts?

Is there a timeline for when you want to merge this?

2color:
> do you want me to fix the conflicts?

@0x6431346e That would be greatly appreciated. 

This is of high priority right now, so the idea is to get it merged in the coming days.

packages/aragon-wrapper/src/index.js
```diff
@@ -1,5 +1,9 @@
 // Externals
-import { ReplaySubject, Subject, BehaviorSubject, Observable } from 'rxjs/Rx'
+import { ReplaySubject, Subject, BehaviorSubject, combineLatest, merge } from 'rxjs'
+import {
+  map, startWith, scan, tap, publishReplay, switchMap, flatMap, filter, take,
+  debounceTime, skipWhile
```
sohkai: Typically we'd put each of these imports on their own line, alphabeticalized

packages/aragon-wrapper/src/index.js
```diff
@@ -596,14 +608,14 @@ export default class Aragon {
...
+      map((apps) => {
         const app = apps.find((app) => addressesEqual(app.proxyAddress, proxyAddress))
         // TODO: handle undefined (no proxy found), otherwise when calling app.proxyAddress next, it will throw
         return makeProxyFromABI(app.proxyAddress, app.abi, this.web3)
-      })
-      .take(1)
-      .toPromise()
+      }),
+      take(1)
+    ).toPromise()
```
sohkai: Maybe we could take this entire bit out of the observables.

What we currently do:

1. From `apps` observable, map each emission
1. When mapping each emission, get an `AppProxy` out
1. Oh wait—we only want the first emission

What do you think about changing it to:

1. Get first emission of `apps` observable (since it's `publishReplay()`'d, it's the current value) and turn it into a promise that resolves with an array
1. Do a normal operations with the array to `find()` and get an `AppProxy`

This way it'd be much easier to emit errors from here in the future as well.

packages/aragon-wrapper/src/index.js
```diff
@@ -669,9 +682,9 @@ export default class Aragon {
    * @return {Promise<Array<string>>} An array of addresses
    */
   getAccounts () {
-    return this.accounts
-      .take(1)
-      .toPromise()
+    return this.accounts.pipe(
+      take(1)
```
sohkai: We should change all `take(1)` to `first()`

packages/aragon-wrapper/src/index.js
```diff
@@ -1207,7 +1221,7 @@ export default class Aragon {
       } catch (_) { }
     }
 
-    const forwarders = await this.forwarders.take(1).toPromise().then(
+    let forwarders = await this.forwarders.pipe(take(1)).toPromise().then(
```
sohkai: Can this be left `const` (nothing else seems to have changed)?

packages/aragon-wrapper/src/index.test.js
```diff
@@ -616,42 +593,36 @@ test('should throw if no ABI is found, when calculating the transaction path', a
...
-    ])
+    }
   })
+  instance.forwarders = of([
+    {
+      appId: 'forwarderA',
+      proxyAddress: '0x999'
+    }
+  ])
+  instance.apps = of([
```
sohkai: Small note: `from()` and `of()` will create Observables that _end_, whereas our previous creation methods created Observables that didn't. This latter behaviour is actually more realistic most of the time; our Observables never end because Ethereum never ends.

This isn't a big deal, and I think we should always be programming Observables with the assumption that they can end anyway (if we ever handle network error states, etc).

packages/aragon-wrapper/src/rpc/handlers/cache.js
```diff
@@ -1,6 +1,8 @@
 export default function (request, proxy, wrapper) {
   const cacheKey = `${proxy.address}.${request.params[1]}`
   if (request.params[0] === 'get') {
+    // FIXME Isn't this obsolete?
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/observe
     return wrapper.cache.observe(cacheKey)
```
sohkai: I think we can remove this comment; we're using `cache.observe()`, not the "standard" `observe`.

packages/aragon-wrapper/src/rpc/handlers/index.js
```diff
@@ -14,29 +15,29 @@ export function createResponse ({ request: { id } }, { error, value = null, kind
...
      * - if the promise resolves emit a Notification of kind 'N' (next) with a value property AND
      * another one of kind 'C' (complete) which we should filter out
      */
-    ({ request, proxy, wrapper }) => Observable.from(
-      handler(request, proxy, wrapper)
-    ).materialize(),
-    createResponse
-  ).filter(
+    mergeMap(
+      ({ request, proxy, wrapper }) => from(handler(request, proxy, wrapper)).pipe(materialize()),
```
sohkai: Let's indent this a bit to make it more readable

2color:
Strange, since the last two commits the package size has increased, hence the build failing. 

Investigating whether it's related to the absence of a `package-lock.json`

2color:
Update: When deleting the `node_modules` and automatically created `package-lock.json`, running an `npm install` and then running size-limit it exceeds the set limit. I'll temporarily increase the limit so we're not blocked by this. 

In addition I'm [double checking](https://aragon.chat/channel/dev-frontend?msg=TaGowyzEk5K3a4Xye) what's the exact error that prevents us from adding a `package-lock.json`

sohkai:
The size increasing that much is quite surprising... we might want to dig into a visualizer to see exactly what's causing that increase (especially since RxJS should have a much smaller footprint now...).

2color:
Strangely I was getting a different size in the UI from the console.
In the console: 
```
  @aragon/wrapper
  Package size: 820.24 KB
  Size limit:   850 KB
```

in the UI `716 KB`
screenshot 2019-03-05 at 01 00 21

sohkai:
Looks like `bn.js` moved versions underneath us 😨. `ethereum-ens` updated the web3 dependency underneath us, and so we have a bunch of mismatches (and duplications) with `bn.js`.

PR #126, opened 2018-07-14
Add support for { from: '' } param in call method

### LINKED ISSUE #125: Add a sender address parameter to AragonApp.call()
The `call(...)` function uses a default address as the sender for every call. Even though it works well in the majority of cases, it makes it pretty hard to build contract methods whose behavior changes depending on `msg.sender`. 

It would be a nice addition to be able to specify a `{ from: '...' }` parameter in `AragonApp.call()` to let the wrapper know which address we want to use.

leftab:
This closes #125 
With this change, a developer can specify the call's sender address like this:

```javascript 
app.call('foo', 'bar', { from: '0x5f6f7e8cc7346a11ca2def8f827b7a0b612c56a1' })
```

### DIFF
--- a/docs/APP.md
+++ b/docs/APP.md
@@ -230,7 +230,7 @@ Perform a read-only call on the app's smart contract.
 
 1. `method` (`String`): The name of the method to call.
 2. `...params` (*arguments*): An optional variadic number of parameters.
-3. `options` (`Object`): Call options (Optional). See [web3 doc](https://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#id16) for more details. 
+3. `options` (`Object`): Call options (Optional). See the [web3.js doc](https://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#id16) for more details. 
 
 **Returns**
 
--- a/docs/APP.md
+++ b/docs/APP.md
@@ -230,6 +230,10 @@ Perform a read-only call on the app's smart contract.
 
 1. `method` (`String`): The name of the method to call.
 2. `...params` (*arguments*): An optional variadic number of parameters.
+3. `options` (`Object`): Call options (Optional).
+    * `from` (`String`): The address the call “transaction” should be made from (Optional).
+    * `gasPrice` (`String`): The gas price in wei to use for this call “transaction” (Optional).
+    * `gas` (`Number`): The maximum gas provided for this call “transaction” (Optional).
 
 **Returns**
 
--- a/docs/APP.md
+++ b/docs/APP.md
@@ -230,10 +230,7 @@ Perform a read-only call on the app's smart contract.
 
 1. `method` (`String`): The name of the method to call.
 2. `...params` (*arguments*): An optional variadic number of parameters.
-3. `options` (`Object`): Call options (Optional).
-    * `from` (`String`): The address the call “transaction” should be made from (Optional).
-    * `gasPrice` (`String`): The gas price in wei to use for this call “transaction” (Optional).
-    * `gas` (`Number`): The maximum gas provided for this call “transaction” (Optional).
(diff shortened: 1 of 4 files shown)


packages/aragon-wrapper/src/core/proxy/index.js
```diff
@@ -47,7 +47,17 @@ export default class Proxy {
...
 
-    return this.contract.methods[method](...params).call()
+    const { from } = params[params.length - 1]
+
+    if (typeof lastParam === 'object' && lastParam !== null) {
+      // Add `from` to call params only if it's defined, can be easily extended for more params
+      return this.contract.methods[method](...params.slice(0,-1)).call({ 
+        ...(from && { from }) 
+      })
+    }
```
sohkai: A few nitpicks on the above block:

- `lastParam` doesn't seem to be defined
- `...params.slice(0,-1)` doesn't need the spread, since it makes a new array already
	- I'm not sure why it was being spread in the original case, and we can probably remove that
- We can probably just pass in the entire `lastParam` to `call()` so any of the [optional parameters](http://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#id16) are supported
- Would be great to document here: https://github.com/aragon/aragon.js/blob/master/docs/APP.md#call

packages/aragon-wrapper/src/core/proxy/index.js
```diff
@@ -47,7 +47,17 @@ export default class Proxy {
...
+    else {
+      return this.contract.methods[method](...params).call()
```
sohkai: This bit doesn't need the `else`, since the above block will return if executed.

max-cote:
Thanks a lot for the feedback @sohkai ! I should be able to update the code in a couple of days.

leftab:
Ok sorry for the hude delay! I guess we were all a little bit too optimistic about what we would be able to do in the first week of our project! haha

So thanks again for this thoughtful feedback! Here is what I found:
   - You are totally right! :smile: My bad, too much refactoring. This should be fixed.
   - After playing a little bit with web3, the spread is indeed necessary in some cases. e.g. calling a method that doesn't take any parameters with an empty array will result in the following error: `Invalid number of parameters for "..." . Got 1 expected 0!`
   - Agree! Done!
   - Also done :o)
   - And now that the lastParam is back, the else is useful again hehe

packages/aragon-wrapper/src/core/proxy/index.js
```diff
@@ -46,8 +46,15 @@ export default class Proxy {
...
     }
+    
+    const lastParam = params[params.length - 1] 
 
-    return this.contract.methods[method](...params).call()
+    if (typeof lastParam === 'object' && lastParam !== null) {
+      return this.contract.methods[method](...params.slice(0,-1)).call(lastParam)
+    }
+    else {
+      return this.contract.methods[method](...params).call()
```
sohkai: This `else` isn't necessary, since if the previous condition was true, it'd never reach here :).

What do you think about making this a ternary? I like it because the alignment makes it basically a diff of the two options.

```js
return (typeof lastParam === 'object' && lastParam !== null)
  ? this.contract.methods[method](...params.slice(0,-1)).call(lastParam)
  : this.contract.methods[method](...params).call()
```

docs/APP.md
```diff
@@ -230,6 +230,10 @@ Perform a read-only call on the app's smart contract.
 
 1. `method` (`String`): The name of the method to call.
 2. `...params` (*arguments*): An optional variadic number of parameters.
+3. `options` (`Object`): Call options (Optional).
```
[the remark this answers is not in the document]
leftab: Done :slightly_smiling_face:

packages/aragon-wrapper/src/core/proxy/index.js
leftab: Oh sorry I didn't understand what you meant the other day. Yeah I have to say it is cleaner that way! :)

sohkai:
@leftab Thanks again :pray:!

Issue #18, opened 2017-10-20
Increase test coverage


onbjerg:

0x6431346e:
Related: #161  #147  #129

PR #129, opened 2018-08-07
Add unit tests to aragon-messenger

0x6431346e:
This is work in progress, part of #18 , but before I continue, which mocking library to use? sinon.js?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ notifications:
   email: false
 node_js:
   - '8'
+before_install:
+  - npm install --global npm
 before_script:
   - npm prune
 script:
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,2881 @@
+{
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "JSONStream": {
+      "version": "1.3.4",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.4.tgz",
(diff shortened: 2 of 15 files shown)


sohkai:
I definitely prefer sinon by default, unless there's a strong reason to use a different mocking library?

0x6431346e:
Finished with `aragon-messenger`, if this makes sense we can merge this and I will open different PRs for the other packages.

I've also added `proxyquire` because I had some trouble mocking `./jsonrpc` from `index.js`.

Quazia:
Looks good to me, a few minor things that might make sense:
I would either 
a. alter travis to only run tests on the package that has tests implemented 
or 
b. get rid of the test file at ``packages/aragon-wrapper/end/index.test.js`` which seems to be responsible for the CI failing. I'd go with the latter.

Also, I think it makes sense to remove ``packages/aragon-messenger/test/placeholder.js`` so there's not a random empty todo test popping up.

0x6431346e:
From what I can tell, the builds are failing on travis because the `websocket` module is missing, although it it should be there.
Upgrading to `npm 6.4.0` seems to fix this issue, so I've added an extra step in `travis.yml` to make sure npm is the latest.
This won't worked until the PR is merged though, because travis is using the config file from master instead of the one from the branch. :thinking:

Quazia:
Can you remove the ``packages/aragon-messenger/test/placeholder.js`` file as well. And the ``packages/aragon-wrapper/ens/index.test.js`` file unless you think thats worth keeping, I'll leave that one up to you if you think we should keep but I think removing at least the placeholder makes sense.

0x6431346e:
Yep, makes sense.
I think it's worth keeping the `ens/index.test.js` file for now. I will take another look at it next, when I'll write unit tests for the wrapper and the client.

PR #131, opened 2018-08-10
Support optional transaction parameters in intents

izqui:
This allows Aragon apps to specify custom transaction values such as `gas`, `value` or `gasPrice` when creating an intent.

### DIFF
--- a/docs/APP.md
+++ b/docs/APP.md
@@ -56,7 +56,7 @@ const app = new AragonApp()
 app.increment(1)
 ```
 
-You can also pass an optional object after all the required function arguments specify some values that will be sent in the transaction. They are the same values that can be passed to `web3.eth.sendTransaction()` and can be checked in this [web3.js document](https://web3js.readthedocs.io/en/1.0/web3-eth.html#id62).
+You can also pass an optional object after all the required function arguments to specify some values that will be sent in the transaction. They are the same values that can be passed to `web3.eth.sendTransaction()` and can be checked in this [web3.js document](https://web3js.readthedocs.io/en/1.0/web3-eth.html#id62).
 
 ```js
 app.increment(1, { gas: 200000, gasPrice: 80000000 }) // careful: hardcoding the gas limit could break your app!
--- a/docs/APP.md
+++ b/docs/APP.md
@@ -46,15 +46,27 @@ The app communicates with the wrapper using a messaging provider. The default pr
 
 To send an intent to the wrapper (i.e. invoke a method on your smart contract), simply call it on the instance of this class as if it was a JavaScript function.
 
-For example, to invoke `increment` on your app's smart contract:
+For example, to execute the `increment` function in your app's smart contract:
 
 ```js
 const app = new AragonApp()
 
 // Sends an intent to the wrapper that we wish to invoke `increment` on our
 // app's smart contract
-app.increment()
+app.increment(1)
 ```
+
+You can also pass an optional object after all the required function arguments specify some values that will be sent in the transaction. They are the same values that can be passed to `web3.eth.sendTransaction()` and can be checked in this [web3.js document](https://web3js.readthedocs.io/en/1.0/web3-eth.html#id62).
+
+```js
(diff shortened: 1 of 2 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -642,17 +642,33 @@ export default class Aragon {
...
+    }
+
+    // If an extra parameter has been provided, it is the transaction options
+    if (methodABI.inputs.length + 1 == params.length && typeof params[params.length - 1] === 'object') {
+      const options = params.pop()
+      transactionOptions = { ...transactionOptions, ...options }
+    }
+
+    if (methodABI.inputs.length != params.length) {
+      throw new Error(`Incorrect number of parameters for function ${methodABI.name}. Expected ${methodABI.input.length}, provided ${params.length}`)
```
Quazia: If the above conditional fails on the typeof statement I could make a case for having a separate error message but it's probably not necessary.

Quazia:
LGTM

docs/APP.md
```diff
@@ -46,16 +46,28 @@ The app communicates with the wrapper using a messaging provider. The default pr
...
 ```js
 const app = new AragonApp()
 
 // Sends an intent to the wrapper that we wish to invoke `increment` on our
 // app's smart contract
-app.increment()
+app.increment(1)
 ```
 
+You can also pass an optional object after all the required function arguments specify some values that will be sent in the transaction. They are the same values that can be passed to `web3.eth.sendTransaction()` and can be checked in this [web3.js document](https://web3js.readthedocs.io/en/1.0/web3-eth.html#id62).
```
sohkai: Typo: function arguments **to** specify

docs/APP.md
```diff
@@ -46,16 +46,28 @@ The app communicates with the wrapper using a messaging provider. The default pr
...
+
+```js
+app.increment(1, { gas: 200000, gasPrice: 80000000 }) // careful: hardcoding the gas limit could break your app!
+```
+
+Some caveats to customizing transaction parameters:
+
+- `from`, `to`, `data`: will be ignored as aragon.js will calculate those.
+- `gas`: The gas amount required for an action will vary depending on how the action ends up being executed (if it goes through a forwarder it will require more gas). Manually specifying the gas limit is not recommended as it could break your app depending on what DAOs it gets installed to.
```
sohkai: In the future, it may be interesting to add something like `minGas` to specify a lower bound on gas. I think the cases where gas estimation get stuck is when it estimates a really low amount of gas so this could avoid that while still allowing for higher cost operations like forwarding.

packages/aragon-wrapper/src/index.js
```diff
@@ -791,6 +777,35 @@ export default class Aragon {
...
       }
 
+      const methodABI = app.abi.find(
+        (method) => method.name === methodName
+      )
+      if (!methodABI) {
+        throw new Error(`${methodName} not found on ABI for ${destination}`)
+      }
+
+      let transactionOptions = {
```
sohkai: This declaration needs to be moved to outside the `if` scope, since `createForwarderTransaction()` needs it later

packages/aragon-wrapper/src/index.js
```diff
@@ -791,6 +777,35 @@ export default class Aragon {
...
+      }
+
+      // If an extra parameter has been provided, it is the transaction options if it is an object
+      if (methodABI.inputs.length + 1 == params.length && typeof params[params.length - 1] === 'object') {
+        const options = params.pop()
+        transactionOptions = { ...transactionOptions, ...options }
+      }
+
+      if (methodABI.inputs.length != params.length) {
+        throw new Error(`Incorrect number of parameters for function ${methodABI.name}. Expected ${methodABI.input.length}, provided ${params.length}`)
```
sohkai: `web3.eth.abi.encodeFunctionCall()` will throw in this case, so we might not need an explicit check here.

docs/APP.md
izqui: Great idea. Created an issue so we don't forget about it https://github.com/aragon/aragon.js/issues/150. Also thinking about whether we should always

PR #137, opened 2018-08-24
Add permission manager ACL representation

### LINKED ISSUE #135: Add permission managers to ACL representation
Currently the aragon.js representation of the ACL is missing the permission managers. We need to watch for the `SetPermissionManager` event and save the data.

izqui:
Close #135 

Changes how data is stored in the permissions object, rather than `app -> role -> [ entities ]` it now stores an object `app -> role -> { allowedEntities: [ entities ], manager }`

Edit: initially also tracked the permission param hashes, but reverted the changes given that it would have required updating the ABI to the latest aragonOS.

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -139,7 +139,7 @@ export default class Aragon {
     // Set up permissions observable
 
     // Permissions Object:
-    // app -> role -> { manager, allowedEntities -> [ entities with permission ], paramHashes -> { entity -> param hash } } 
+    // app -> role -> { manager, allowedEntities -> [ entities with permission ] }
     this.permissions = Observable.merge(...aclObservables)
       .scan((permissions, event) => {
         const eventData = event.returnValues
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -145,7 +145,7 @@ export default class Aragon {
         const eventData = event.returnValues
         const baseKey = `${eventData.app}.${eventData.role}`
 
-        if (event.event == SET_PERMISSION_EVENT) {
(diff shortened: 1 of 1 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -126,25 +126,39 @@ export default class Aragon {
...
+    ].map(event =>
+      this.aclProxy.events(event)
+    )
+
     // Set up permissions observable
-    this.permissions = this.aclProxy.events('SetPermission')
-      .pluck('returnValues')
+
+    // Permissions Object:
+    // app -> role -> { manager, allowedEntities -> [ entities with permission ], paramHashes -> { entity -> param hash } }
```
sohkai: Should update this comment, now that the `paramHashes` have been reverted.

packages/aragon-wrapper/src/index.js
```diff
@@ -126,25 +126,39 @@ export default class Aragon {
...
-          `${event.app}.${event.role}`,
-          []
-        )
+        const eventData = event.returnValues
+        const baseKey = `${eventData.app}.${eventData.role}`
 
-        const newPermissionsForRole = event.allowed
-          ? currentPermissionsForRole.concat(event.entity)
-          : currentPermissionsForRole.filter((entity) => entity !== event.entity)
+        if (event.event == SET_PERMISSION_EVENT) {
```
sohkai: Best to use the `===` by default :).

sohkai:
> Edit: initially also tracked the permission param hashes, but reverted the changes given that it would have required updating the ABI to the latest aragonOS.

Can we do this now that the ABI's been switched in #138?

izqui:
> Can we do this now that the ABI's been switched in #138?

I think we should do it in a separate PR to support permission params entirely and not only save the hash.

PR #138, opened 2018-08-26
Implements ACL intents

### LINKED ISSUE #136: Transaction pathing support for ACL writes
Special case of #134 

The `calculateTransactionPath` function can be aided to find a transaction path to a function without a role by providing the address of the last forwarder. This will help transaction pathing for ACL functions as the last forwarder will always be the permission manager (requires #135) 

Transaction pathing to `grantPermission`, `revokePermission`, `setPermissionManager` and `removePermissionManager` require this.

izqui:
Builds on #137 

- Modified `calculateTransactionPath` function so if a `finalForwarder` address is provided, it will always try to forward using that forwarder. In the case of ACL actions that don't have a role, but need the permission manager to perform the action, it passes the permission manager as the `finalForwarder`. Closes #136.
- Added an artifact file for the `ACL` generated locally with the aid of the CLI. This is needed because the ACL app is not currently being published as a regular app.

`wrapper.performACLIntent(method, params)` is used in the same way that intents are used inside regular apps. It will perform a special case of transaction pathing taking permission managers into account (except for `createPermission` which uses a normal role) and then add the transaction.
- `method` is the function name (i.e. `grantPermission`)
- `params` is an array of the parameters (i.e.`[entity, app, role]`)

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -133,7 +133,7 @@ export default class Aragon {
 
     const aclObservables = [
       SET_PERMISSION_EVENT,
-      CHANGE_PERMISSION_MANAGER_EVENT,
+      CHANGE_PERMISSION_MANAGER_EVENT
     ].map(event =>
       this.aclProxy.events(event)
     )
@@ -141,7 +141,7 @@ export default class Aragon {
     // Set up permissions observable
 
     // Permissions Object:
-    // app -> role -> { manager, allowedEntities -> [ entities with permission ], paramHashes -> { entity -> param hash } } 
+    // app -> role -> { manager, allowedEntities -> [ entities with permission ], paramHashes -> { entity -> param hash } }
     this.permissions = Observable.merge(...aclObservables)
       .scan((permissions, event) => {
         const eventData = event.returnValues
@@ -149,7 +149,7 @@ export default class Aragon {
 
         if (event.event == SET_PERMISSION_EVENT) {
           const key = `${baseKey}.allowedEntities`
-          const currentPermissionsForRole = dotprop.get(permissions, key , [])
+          const currentPermissionsForRole = dotprop.get(permissions, key, [])
 
           const newPermissionsForRole = eventData.allowed
             ? currentPermissionsForRole.concat(eventData.entity)
@@ -497,7 +497,7 @@ export default class Aragon {
    * @param {Array<Object>} An array of Ethereum transactions that describe each step in the path
    * @return {Promise<string>} transaction hash
    */
-  performTransactionPath(transactionPath) {
+  performTransactionPath (transactionPath) {
     return new Promise((resolve, reject) => {
       this.transactions.next({
         transaction: transactionPath[0],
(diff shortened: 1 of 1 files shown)


packages/aragon-wrapper/src/core/aragonOS/index.js
```diff
@@ -0,0 +1,22 @@
+import { hash as namehash } from 'eth-ens-namehash'
```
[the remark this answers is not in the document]
izqui: 😅 forgot to commit the file

packages/aragon-wrapper/src/index.js
```diff
@@ -218,7 +234,7 @@ export default class Aragon {
           apps.map(async (app) => Object.assign(
             app,
             await this.apm.getLatestVersionForContract(app.appId, app.codeAddress)
-              .catch(() => ({}))
+              .catch(() => getAragonOSAppInfo(app.appId)) // for unpublished apps we check local mapping
```
sohkai: @bpierre this should give us the ACL and its artifact as an app in the `apps` observable, so we can remove the some of the hardcoded roles in aragon/aragon :)

izqui: Yep, I was also thinking whether it would be beneficial to add the kernel to the app list as well.

packages/aragon-wrapper/src/index.js
```diff
@@ -477,6 +493,85 @@ export default class Aragon {
...
+      const appIndex = inputNames.indexOf('_app')
+      const roleIndex = inputNames.indexOf('_role')
+
+      if (appIndex == -1 || roleIndex == -1) {
+        throw new Error(`Method ${method} doesn't take _app and _role as input. Permission manager cannot be found.`)
+      }
+
+      const manager = await this.getPermissionManager(params[appIndex], params[roleIndex])
+
+      const path = await this.getTransactionPath(aclAddr, method, params, manager)
```
sohkai: We should double check that the `manager` is a forwarder.

packages/aragon-wrapper/src/index.js
sohkai: Was thinking about that too... not sure if it really belongs in the same list as it's not really an "app", but that would likely be the most convenient.

Another alternative is a `kernel` observable that returns the artifact and address of the kernel.

packages/aragon-wrapper/src/index.js
```diff
@@ -477,6 +493,92 @@ export default class Aragon {
...
+   * Performs an action on the ACL using transaction pathing
+   *
+   * @param {string} method
+   * @param {Array<*>} params
+   * @return {Promise<string>} transaction hash
+   */
+  async performACLIntent (method, params) {
+    const aclAddr = this.aclProxy.address
+
+    if (method === 'createPermission') {
```
sohkai: Would it be feasible to filter the acl's artifact for methods that have roles attached instead of hardcoding names?

izqui: Good catch, we actually can, adding it

PR #140, opened 2018-08-28
Wrapper: new API for init() and initAccounts()

### LINKED ISSUE #128: Wrapper: init() should only fetch accounts from the web3 instance if requested
Doing this by default leads to weird situations if a client is set up to use two `web3` providers: one for fetching events and another as the actual wallet.

If the wallet provider is locked, but the events one isn't, there will be accounts associated with the user that he can't actually use.

Getting the accounts by default should only be done if requested via an option.

juslar:
Fix #128, implements API in https://github.com/aragon/aragon.js/pull/140#discussion_r213603420.


docs/WRAPPER.md
```diff
@@ -14,18 +14,22 @@
...
 
 Initialise the wrapper.
 
+### **Parameters**
+
+-   `accounts` **[Array](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array)&lt;string>**
+      An optional array of accounts that the user controls
+-   `options` **[Object](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Object)** 
+      An optional options object that contains 'withAccounts' that specifies whether or not we should fetch accounts from the Web3 instance.
+      Ex: ```aragon.init(["0xbeefbeef"], { withAccounts: true })```
```
Quazia: This example doesn't make sense with the way you have it set up. Currently, it will _only_ pull accounts if the account param is null, it might make sense to have the accounts param and the web3 accounts combined if it's called in the way shown in this example.

sohkai: The two can't be combined into a single parameter because there may be times where we'd like to not provide an 
account list but still not fetch accounts from web3.

Given that users should know if they'd like to fetch accounts or provide their own by the time they init, a breaking API change would be to turn the entire argument list into a single options object with `accounts` and `fetchAccountsFromConnectedWeb3` options. If both are provided, the web3 accounts could be concatenated with the provided accounts.

juslar:
Thanks! I made the change you requested @sohkai

sohkai:
@juslar Sorry for the delay... give us a bit to get back to this and review it :).

sohkai:
@juslar I've updated this PR and modified it to the API in https://github.com/aragon/aragon.js/pull/140#discussion_r213603420 now that we can break things with a new major version of aragon.js :).

I think the new API's a lot more clear, as it's more declarative, and also more easily lets us expand the options later if we need to in `.init()` :).

Thanks for helping get this started!

packages/aragon-wrapper/src/index.js
```diff
@@ -79,11 +79,11 @@ const appInfoCache = {}
...
  *
- * // Initialises the wrapper and logs the installed apps
- * aragon.init(() => {
- *   aragon.apps.subscribe(
- *     (apps) => console.log(apps)
- *   )
+ * // Initialises the wrapper
+ * await aragon.init({
+ *   accounts: {
+ *     provided: ["0xbeefdead", "0xbeefbeef"]
```
izqui:
```suggestion
 *     providedAccounts: ["0xbeefdead", "0xbeefbeef"]
```

PR #147, opened 2018-09-03
Add unit tests to aragon-wrapper

0x6431346e:
This is about adding some unit tests for the wrapper. I just focused on the code that depends on `RxJS`, so we can merge #123 and make sure nothing breaks.

There are a few more functions and files to test though.

TODO:
- [x] index.js/runApp
- [x] index.js/getPermissionManager
- [x] index.js/getApp
- [x] index.js/calculateTransactionPath (Partly)
- [x] rpc/handlers/external.js
- [x] rpc/handlers/index

### DIFF
--- a/packages/aragon-wrapper/src/cache/index.js
+++ b/packages/aragon-wrapper/src/cache/index.js
@@ -40,14 +40,15 @@ export default class Cache {
     return `${this.prefix}.${key}`
   }
 
-  set (key, value) {
+  async set (key, value) {
     // Some lowdb adapters are synchronous while others are asynchronous so
     // let's always wrap it in a promise
-    return Promise.resolve(this.db.set(
+    await Promise.resolve(this.db.set(
       this.getCacheKeyPath(key),
       value
     ).write())
-      .then(() => this.changes.next({ key, value }))
+     
+    this.changes.next({ key, value })
   }
 
   get (key, defaultValue) {
@@ -71,10 +72,9 @@ export default class Cache {
    * @return {Observable}
    */
   observe (key, defaultValue) {
-    return this.changes.filter(
-      (change) => change.key === key
-    ).map(
-      (change) => change.value
-    ).startWith(this.get(key, defaultValue))
+    return this.changes
+      .filter((change) => change.key === key)
+      .map((change) => change.value)
+      .startWith(this.get(key, defaultValue))
   }
 }
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -1,5 +1,5 @@
 // Externals
-import { Subject, BehaviorSubject, Observable } from 'rxjs/Rx'
+import { ReplaySubject, Subject, BehaviorSubject, Observable } from 'rxjs/Rx'
 import { isBefore } from 'date-fns'
 import uuidv4 from 'uuid/v4'
 import Web3 from 'web3'
@@ -123,7 +123,9 @@ export default class Aragon {
    * @param {?Array<string>} [accounts=null] An optional array of accounts that the user controls
    * @return {Promise<void>}
    */
-  async initAccounts (accounts = null) {
+  async initAccounts (accounts) {
+    this.accounts = new ReplaySubject(1)
+
     if (accounts === null) {
       accounts = await this.web3.eth.getAccounts()
     }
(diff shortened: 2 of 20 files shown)


packages/aragon-wrapper/src/rpc/handlers/index.js
```diff
@@ -1,6 +1,6 @@
 import { Observable } from 'rxjs/Rx'
 
-export function createResponse ({ request: { id } }, { error, value = null }) {
```
0x6431346e: I removed the default assignment here from `null` because it was causing duplicates to be emitted.
(As long as they remain `undefined` they will get filtered in `createRequestHandler`)

izqui:
When #160 lands it will be important to test that too.

packages/aragon-wrapper/src/rpc/handlers/index.js
sohkai: I think the reason they're `null` is because JSON-RPC can't transmit `undefined` values over the wire.

packages/aragon-wrapper/src/index.js
```diff
@@ -111,9 +112,7 @@ export default class Aragon {
    * @param {?Array<string>} [accounts=null] An optional array of accounts that the user controls
    * @return {Promise<void>}
    */
-  async initAccounts (accounts) {
-    this.accounts = new ReplaySubject(1)
```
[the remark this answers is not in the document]
0x6431346e: We don't return the observable, we convert it into a promise, so we only use `ReplaySubject` internally:
```
  getAccounts () {
    return this.accounts
      .take(1)
      .toPromise()
  }
```
Anyway, I reverted it back since it's a silly change and not important to this PR.

packages/aragon-wrapper/src/rpc/handlers/index.js
0x6431346e: Oops, nice catch!
I fixed the duplicates by explicitly checking `kind === 'C'`.
Also added a test check for `undefined` results that should be converted to `null`.

jvluso:
This looks awesome. Thank you for adding these. The tests are not all passing when I run this. I get the following result:

```
> @aragon/wrapper@2.0.0-beta.44 test /home/jeremy/aragon.js/packages/aragon-wrapper
> nyc --silent --all ava


  39 passed
  1 failed

  dist › index › should init the ACL correctly

  /home/jeremy/aragon.js/packages/aragon-wrapper/dist/index.js:56

   55: Object.defineProperty(exports, 'ensResolve', {
   56:   enumerable: true,                           
   57:   get: function get() {                       

  Rejected promise returned by test. Reason:

  TypeError {
    message: 'Cannot read property \'events\' of undefined',
  }

  dist/index.js:56:426
  dist/index.js:56:337
npm ERR! Test failed.  See above for more details.
```

Can you look into this?

0x6431346e:
>   dist › index › should init the ACL correctly
> 
>   /home/jeremy/aragon.js/packages/aragon-wrapper/dist/index.js:56

My guess is that some old build files are at fault. :thinking: 

Could you try now? I've updated the test commands to include only files matching `src/**/*.test.js`.

jvluso:
how many tests should there be? There were 40 earlier when it failed, but now there are 5.

packages/aragon-wrapper/src/cache/index.js
```diff
@@ -40,14 +40,15 @@ export default class Cache {
...
     // Some lowdb adapters are synchronous while others are asynchronous so
     // let's always wrap it in a promise
-    return Promise.resolve(this.db.set(
+    await Promise.resolve(this.db.set(
       this.getCacheKeyPath(key),
       value
     ).write())
-      .then(() => this.changes.next({ key, value }))
+     
+    this.changes.next({ key, value })
```
sohkai: @0x6431346e I don't think this has broken anything, but I think this has changed the API of `set()`. We should return the value here to keep it the same as before.

0x6431346e: Seems to return `undefined` though, the docs also say `public next(value: T): void`  :relieved: 
http://reactivex.io/rxjs/class/es6/MiscJSDoc.js~ObserverDoc.html#instance-method-next

sohkai: Oh good catch :). I just assumed it would return the observable.

PR #148, opened 2018-09-03
wrapper acl: prevent allowed entities from being duplicated

izqui:
The ACL smart contract doesn't fail if granting an already existing permission, and it emits an extra `SetPermission` event. When doing so, the ACL representation of aragon.js was showing some entities  multiple times.


Quazia:
Looks good. As a note maybe add a test for this behavior in the wrapper testing @0x6431346e

izqui:
@sohkai Manually granting permissions repeatedly to the same entity

PR #151, opened 2018-09-04
31 emit rpc errors

### LINKED ISSUE #31: Emit errors in RPC messenger when an error response is received
Whenever a response is received in the RPC messenger (`src/rpc/Messenger.js`) and it is an error, then an error should be emitted on the observable as opposed to just passing it through as a normal response.

The responses and errors conform to the JSONRPC 2.0 spec.

Quazia:
Was initially confused about where this should be emitted but I think they're being emitted on the observable correctly now.

Fixes #31.

### DIFF
--- a/packages/aragon-messenger/src/index.test.js
+++ b/packages/aragon-messenger/src/index.test.js
@@ -1,0 +1,161 @@
+import test from 'ava'
+import sinon from 'sinon'
+import proxyquire from 'proxyquire'
+import { Subject } from 'rxjs'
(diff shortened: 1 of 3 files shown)


sohkai:
Have to think about the messages flow, and where we'll be catching these errors to see if this is the right place to handle this.

At the very least, if it isn't, we clearly need to be passing more information out from `.message()`.

sohkai:
Superseded by #277.

PR #152, opened 2018-09-04
64 error on bad dao

### LINKED ISSUE #64: Wrapper should error if the provided address isn't a DAO
Otherwise, the app breaks because `this.kernelProxy.call('acl')` returns `0x0` and web3 complains it can't make an ACL out of that.

@bpierre we should detect the thrown error and display a "No organization found" screen or similar.

Quazia:
Just doing the basic check to see if there's an ACL since there isn't a helper function for this yet, may want to open a new issue for if/when the isDAO function exists.

Fixes #64.

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -77,8 +77,14 @@ export default class Aragon {
       provider: options.provider
     }))
 
+
     // Set up the kernel proxy
     this.kernelProxy = makeProxy(daoAddress, 'Kernel', this.web3)
+    
+    // Check if address is kernal
+    const aclAddress = await this.kernelProxy.call('acl')
+    if ( aclAddress === "0x0")
+      throw Error("Provided Address is not a Dao")
 
     // Set up cache
     this.cache = new Cache(daoAddress)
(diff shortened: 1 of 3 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -77,8 +77,14 @@ export default class Aragon {
       provider: options.provider
     }))
 
+
     // Set up the kernel proxy
     this.kernelProxy = makeProxy(daoAddress, 'Kernel', this.web3)
+    
+    // Check if address is kernal
```
PascalPrecht: kernal => kernel?

sohkai:
@Quazia I've merged this PR again master and moved the check to be in `init()` instead.

This also gave an opportunity to allow a user to use `initAcl()` with a non-default own ACL instance, if they wanted to (no real use case for this yet).

packages/aragon-wrapper/src/index.test.js
```diff
@@ -3,22 +3,57 @@ import sinon from 'sinon'
...
+    makeAddressMapProxy: sinon.fake.returns({}),
+    makeProxy: sinon.stub(),
+    addressesEqual: Object.is
+  }
+  const Aragon = proxyquire.noCallThru().load('./index', {
+    '@aragon/messenger': messengerConstructorStub,
+    './utils': utilsStub
+  }).default
+
+  t.context = {
```
sohkai: Moving all of this into context means they get recreated on each test, so the stubs don't share any history across tests :).

PR #153, opened 2018-09-04
#130 tweak wrapper options

### LINKED ISSUE #130: Wrapper: options object is an option parameter, but an error is thrown if it is omitted.
If `options.apm` is omitted, or the options object is omitted entirely an error is thrown. Additionally, `options.apm` is undocumented: only `options.provider` and `options.ensRegistryAddress` are listed in the documentation.

See line 75 in aragon-wrapper/index.js
`cov_2bk7erqz0c.s[64]++;this.apm=(0,_apm2.default)(this.web3,(0,_assign2.default)(options.apm,{ensRegistryAddress:options.ensRegistryAddress,provider:options.provi
der}));// Set up the kernel proxy
TypeError: Cannot convert undefined or null to object`

Quazia:
This should fix the error that was occurring with issue 130. It might make sense to change the variable names but I think they're probably better as is since there may be other options added at a later time.

Fixes #130.

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -132,9 +132,18 @@ export default class Aragon {
     }
     options = Object.assign(defaultOptions, options)
 
+    // Check for required options
+    if(options.apm === undefined){
+      options.apm = []
+    }
+
+    if(options.ensRegistryAddress === undefined){
+      options.ensRegistryAddress = []
+    }
+
     // Set up Web3
     this.web3 = new Web3(options.provider)
-
+    
     // Set up APM
     this.apm = apm(this.web3, options.apm)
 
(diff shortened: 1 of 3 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -68,9 +68,18 @@ export default class Aragon {
     }
     options = Object.assign(defaultOptions, options)
 
+    // Check for required options
+    if(options.apm === undefined){
+      options.apm = []
+    }
+
+    if(options.ensRegistryAddress === undefined){
+      options.ensRegistryAddress = []
```
sohkai: I think both of these could be defined in the `defaultOptions` object?

As well, I think `options.apm` should default to an empty object, and `options.ensRegistryAddress` should default to a string.

PR #158, opened 2018-09-12
feat(Aragon): introduce defaultGasPriceFn option

### LINKED ISSUE #73: Remove hard coded gas prices and allow for getting gas price from an external source

PascalPrecht:
This option can be used to provide a function that returns a default gas
price. It can be async as well, which enables calls to third party APIs such as
ethgasstation.

Example:
```
const app = new Aragon(daoAddress, {
  provider: someProvider,
  defaultGasPriceFn: () => {
     return new Promise(resolve => {
        resolve('foo')
     })
  }
})
```

Closes #73

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -16,7 +16,7 @@ import * as handlers from './rpc/handlers'
 
 // Utilities
 import { CALLSCRIPT_ID, encodeCallScript } from './evmscript'
-import { addressesEqual, makeProxy, makeProxyFromABI } from './utils'
+import { addressesEqual, makeProxy, makeProxyFromABI, getRecommendedGasLimit } from './utils'
 
 import { getAragonOsInternalAppInfo } from './core/aragonOS'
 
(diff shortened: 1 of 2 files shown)


luisivan:
Should we also include an ethgasstation provider just for ease of use? @izqui

packages/aragon-wrapper/src/index.js
```diff
@@ -69,7 +69,10 @@ export const setupTemplates = (
 export default class Aragon {
   constructor (daoAddress, options = {}) {
     const defaultOptions = {
-      provider: detectProvider()
+      provider: detectProvider(),
+      defaultGasPriceFn: () => {
```
izqui: We should add this new option to the documentation of the constructor.

packages/aragon-wrapper/src/index.js
```diff
@@ -69,7 +69,10 @@ export const setupTemplates = (
...
+        return this.web3.utils.toWei('20', 'gwei')
```
izqui: I think the default function return a Promise even if we resolve immediately: `Promise.resolve(...)`

packages/aragon-wrapper/src/index.js
```diff
@@ -757,7 +766,7 @@ export default class Aragon {
    */
   async calculateTransactionPath (sender, destination, methodName, params, finalForwarder) {
     const finalForwarderProvided = this.web3.utils.isAddress(finalForwarder)
-    const defaultGasPrice = this.web3.utils.toWei('20', 'gwei') // TODO: Get from ethgasstation.info or another source for legit gas values
+    const defaultGasPrice = await this.getDefaultGasPrice()
```
izqui: The gas price that we need to send could be a function of the gas limit that is set in the transaction (higher gas limit may require a higher gas price). We could make the gas the transaction will be sent with a parameter to the `defaultGasPrice` function.

packages/aragon-wrapper/src/index.js
PascalPrecht: @izqui can you elaborate on why this is necessary? The function is synchronous by nature and `await` consumes synchronous functions as well.

I will add it if that's what you want, but I don't see the value it in right now :)

packages/aragon-wrapper/src/index.js
PascalPrecht: Yes. I wanted to wait and see if you're good with `defaultGasPriceFn` or if you prefer a different name.

PascalPrecht:
@izqui updated, PTAL

packages/aragon-wrapper/src/index.js
izqui: The reason I commented that was to make it more explicit to someone reading the code that the function was asynchronous, but of course this would be better solved with proper documentation :)

PascalPrecht:
Closing this in favor of https://github.com/aragon/aragon.js/pull/159

Issue #145, opened 2018-08-30
Add provider engine


jvluso:
A provider engine would allow forwarder accounts in a DAO to be used as an account by truffle. It would also allow commands to be sent through multiple DAOs. The provider should be a web3 provider that takes the following arguments:

- Another web3 provider which all requests are sent to for reading the blockchain
- The dao that the app in question is a part of
- The app that all commands will finally be forwarded out of

Adding this to a truffle configuration should allow a user to interact with the rest of the blockchain using truffle commands as if the chosen app was their account. They should be able to publish aragon apps, update aragon apps, and vote in other daos.It should also be possible to chain multiple of these providers together, and still be able to perform all necessary functions.

Quazia:
This might be a lot to do in a single issue, it might make sense to make this new package an epic and then add sub-tasks?

PR #160, opened 2018-09-18
Isolate functionality to find a forwarder to perform an action

jvluso:
This is going to be necessary for #145 and this part of aragon.js keeps changing, so I wanted to get it into the main branch so that I don't have to keep re merging it.

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -949,7 +949,7 @@ export default class Aragon {
       } catch (_) { }
     }
 
-    let forwarders = await this.forwarders.take(1).toPromise().then(
+    const forwarders = await this.forwarders.take(1).toPromise().then(
       (forwarders) => forwarders.map(
         (forwarder) => forwarder.proxyAddress
       )
@@ -971,8 +971,7 @@ export default class Aragon {
         )
     }
 
-    return await this.calculateForwardingPath (sender, destination, directTransaction, forwardersWithPermission)
-
+    return this.calculateForwardingPath(sender, destination, directTransaction, forwardersWithPermission)
   }
 
   /**
@@ -986,14 +985,6 @@ export default class Aragon {
    * @return {Array<Object>} An array of Ethereum transactions that describe each step in the path
    */
   async calculateForwardingPath (sender, destination, directTransaction, forwardersWithPermission) {
-    const permissions = await this.permissions.take(1).toPromise()
-
-    let forwarders = await this.forwarders.take(1).toPromise().then(
-      (forwarders) => forwarders.map(
-        (forwarder) => forwarder.proxyAddress
-      )
-    )
-
     // No forwarders can perform the requested action
     if (forwardersWithPermission.length === 0) {
       return []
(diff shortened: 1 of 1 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -873,6 +873,31 @@ export default class Aragon {
...
+   * that invokes `directTransaction`.
+   *
+   * @param  {string} sender
+   * @param  {string} destination
+   * @param  {Object} directTransaction
+   * @param  {string} [finalForwarder]
+   * @return {Array<Object>} An array of Ethereum transactions that describe each step in the path
+   */
+  async calculateDirectPath (sender, destination, directTransaction, finalForwarder) {
+    const minGasPrice = this.web3.utils.toWei('20', 'gwei')
```
izqui: Please check the new logic for dealing with gas estimations and gas prices that was added in #159, as these changes are basically reverting that.

packages/aragon-wrapper/src/index.js
```diff
@@ -873,6 +873,31 @@ export default class Aragon {
...
+   * Calculate the transaction path for a transaction to `destination`
+   * that invokes `directTransaction`.
+   *
+   * @param  {string} sender
+   * @param  {string} destination
+   * @param  {Object} directTransaction
+   * @param  {string} [finalForwarder]
+   * @return {Array<Object>} An array of Ethereum transactions that describe each step in the path
+   */
+  async calculateDirectPath (sender, destination, directTransaction, finalForwarder) {
```
izqui: The naming of the function is confusing to me, the intent only got clear after our offline conversation. I think `calculateForwardingPath` would be a way more clear name for describing what this function does.

packages/aragon-wrapper/src/index.js
```diff
@@ -873,6 +873,31 @@ export default class Aragon {
...
+    const minGasPrice = this.web3.utils.toWei('20', 'gwei')
+    const permissions = await this.permissions.take(1).toPromise()
+    const app = await this.apps.map(
+      (apps) => apps.find((app) => addressesEqual(app.proxyAddress, destination))
+    ).take(1).toPromise()
+
+    const finalForwarderProvided = this.web3.utils.isAddress(finalForwarder)
+
+    let permissionsForMethod = []
```
izqui: `permissionsForMethod` should be passed as a parameter to the function, otherwise forwarding when `finalForwarder` is not provided (when transaction pathing within the DAO) is going to break.

packages/aragon-wrapper/src/index.js
```diff
@@ -873,6 +873,31 @@ export default class Aragon {
...
+   * @param  {string} sender
+   * @param  {string} destination
+   * @param  {Object} directTransaction
+   * @param  {string} [finalForwarder]
+   * @return {Array<Object>} An array of Ethereum transactions that describe each step in the path
+   */
+  async calculateDirectPath (sender, destination, directTransaction, finalForwarder) {
+    const minGasPrice = this.web3.utils.toWei('20', 'gwei')
+    const permissions = await this.permissions.take(1).toPromise()
+    const app = await this.apps.map(
```
izqui: It doesn't seem that `app` is needed in the function so it can safely be removed

sohkai:
@jvluso Pinging you just to see if you can get the CLA assistant working. It might be necessary to rebase / recommit the first commit (d64faf4).

packages/aragon-wrapper/src/index.js
```diff
@@ -894,6 +889,12 @@ export default class Aragon {
       } catch (_) { }
     }
 
+    let forwarders = await this.forwarders.take(1).toPromise().then(
```
sohkai: What do you think about putting all of this bit into `calculateForwardingPath()`, so that `calculateTransactionPath()` just passes the `finalForwarder`?

It seems like anything forwarder-related should be in `calculateForwardingPath()`, and allow us to remove some of this duplicate logic.

jvluso: Calculating the `forwardersWithPermission` when a `finalForwarder` isn't provided requires the `methodName` to validate permissions. `methodName` isn't (and shouldn't be) passed to `calculateForwardingPath()`.

I also think it makes sense for this to be here because the final forward is the transaction itself, which isn't part of the forwarding path.

sohkai:
@jvluso I've pushed a few commits:

- Merged with master (for `babel@7`)
- Kept the gas option in the forwarding transactions, as it's handled by `applyTransactionGas()`
- A bit of clean up

PR #164, opened 2018-09-27
Transaction pathing: Fix issues introduced with gas estimations

izqui:
Fixes some problems introduced in #159, related to performing gas estimations on intermediate transaction pathing steps.

Also now `gas` is only treated as `minGas` when the transaction is forwarded, when the intent can be performed directly, if `gas` is provided that's the gas that will be sent in the transaction.

Also tested all possible code paths to verify that transaction pathing with multiple hops works as expected:
2018-09-27 at 9 53 18 pm

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -920,7 +920,7 @@ export default class Aragon {
       getAbi('aragon/Forwarder')
     ).methods['forward']
 
-    const createForwarderTransaction = async (forwarderAddress, script) => (
+    const createForwarderTransaction = (forwarderAddress, script) => (
       {
         ...transactionOptions, // Options are overwriten by the values below
         from: sender,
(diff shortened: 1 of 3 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -969,13 +977,15 @@ export default class Aragon {
         if (await this.canForward(forwarder, sender, script)) {
           // The previous forwarder can forward a transaction for this forwarder,
           // and this forwarder can forward for our address, so we have found a path
-          return [await createForwarderTransaction(forwarder, script), ...path]
+          const transaction = createForwarderTransaction(forwarder, script)
+          // `applyTransactionGas` is only done for the transaction that will be executed
+          return [await this.applyTransactionGas(transaction, true), ...path]
```
sohkai: Note that the forwarding transaction here and above can still fail, but given the complexity of making sure this robust, I'm fine with leaving it like this for now. I've added TODO comments to remind us.

izqui:
Thanks for the additions @sohkai!

PR #165, opened 2018-10-01
Wallet Provider

### LINKED ISSUE #145: Add provider engine

jvluso:
Resolves #145 

The wallet provider can be used in the cli to allow Smart Organizations to act as anyone else for any cli.

### DIFF
--- a/packages/aragon-wallet-provider/.babelrc
+++ b/packages/aragon-wallet-provider/.babelrc
@@ -1,0 +1,33 @@
+{
+  "presets": [
+    [
+      "env",
+      {
+        "targets": {
+          "browsers": [
+            ">2%",
+            "last 1 edge versions",
+            "not ie > 0",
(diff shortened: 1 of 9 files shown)


sohkai:
Superseded by efforts in Frame v0.2.

PR #167, opened 2018-10-05
Implement ERC20 approvals with pretransactions

### LINKED ISSUE #166: Add support for ERC20 approvals
ERC20 approvals are a common pattern used for performing actions for that require a token transfer, like depositing in Finance (https://github.com/aragon/aragon-apps/issues/276) or [Staking](https://research.aragon.org/t/staking-locks-spec-v2/217)

Something to explore is whether we can push the second transaction (the call to the contract that will pull the tokens) right after the user signs the first approve transaction (which would be a much better UX). We know for a fact that because of transaction nonces, the second transaction cannot be included into a block before the first one, but signers (and aragon.js' own gas estimation) may not be smart enough to know that the transaction could succeed after the approval transaction is mined.

Also we could support ERC20 approveAndCall and ERC777 transfer hooks if both the token and receiving contract support it.

izqui:
~~Builds on #164~~

Close #166 

The idea is to abstract away from Aragon app developers when tokens need to be approved before performing an intent. In the future, if we start supporting ERC777 or `approveAndCall`, aragon.js could figure out how to send the tokens and execute the desired intent.

Usage:
```
app.deposit(TOKEN_ADDR, { token: { address: TOKEN_ADDR, value: 1 }})
```

I tested it with aragon/aragon just executing the `pretransaction` before the main transaction. When `sendTransaction` returns the transaction hash of the pretransaction, the main transaction can be then sent. From my testing, Metamask doesn't complain about the second transaction failing even before the first one is mined.

Known issues:
- We cannot do the gas estimation for the main transaction as it would fail, so we need to skip it. All workarounds I can think of would add an enormous amount of complexity.
- If the allowance is non-zero, some tokens require setting it to zero before changing it to another non-zero value (to prevent double spending by front-running). This would require us moving to having more than one pretransaction in this case.


sohkai:
Just did some minor clean up in 3d8b09a.

Let's work on the interface later, since this is working quite well ATM :).

PR #168, opened 2018-10-05
Upgrade to babel@7

sohkai:
Upgrades all the packages to use `babel@7`.

Fixes a weird issue with the `babel-runtime` library not working in `babel@7` production environments – if you try to build `aragon/aragon` in production right now (`npm run build`), you get weird errors like the following related to aragon.js:

image

Note that strangely, this doesn't happen in development builds (`npm run start`).

**TODO**:

- [ ] Release new `aragon-messenger` version
- [ ] Bump `aragon-messenger` dependency in `aragon-client` and `aragon-wrapper`
- [ ] Release new `aragon-client` and `aragon-wrapper` versions
- [ ] Test this doesn't break anything

I've tested this by linking `aragon-wrapper` against aragon/aragon and verifying that it resolves the above error. I've also linked `aragon-client` to the Voting app to verify that it hasn't been broken by the upgrade. However, I'm a worried about hidden regressions and it'd be best to play with all the apps a bit once all the releases are done.


packages/aragon-wrapper/.babelrc
```diff
@@ -13,13 +13,14 @@
...
-    ],
-    "stage-2"
+    ]
   ],
   "plugins": [
-    ["inline-json-import", { "extensions": [".json"] }],
-    "transform-flow-strip-types",
-    "transform-runtime"
+    "@babel/plugin-proposal-class-properties",
+    "@babel/plugin-proposal-object-rest-spread",
```
sohkai: I've opted just to include the proposed features we use, rather than use the deprecated stage presets.

packages/aragon-wrapper/package.json
```diff
@@ -28,22 +28,24 @@
...
-    "babel-cli": "^6.26.0",
-    "babel-eslint": "^7.2.1",
-    "babel-plugin-inline-json-import": "^0.2.1",
-    "babel-plugin-transform-flow-strip-types": "^6.22.0",
-    "babel-plugin-transform-runtime": "^6.23.0",
-    "babel-preset-env": "^1.7.0",
-    "babel-preset-minify": "^0.3.0",
-    "babel-preset-stage-2": "^6.24.1",
+    "babel-eslint": "^10.0.1",
+    "babel-plugin-inline-json-import": "0.2.1",
```
sohkai: I've pinned this, as upgrading to `0.3.1` seems to break everything :(

packages/aragon-client/package.json
```diff
@@ -58,7 +60,8 @@
     "babel": "inherit"
   },
   "dependencies": {
-    "rxjs": "^5.5.6",
-    "@aragon/messenger": "^1.0.0-beta.6"
+    "@aragon/messenger": "^1.0.0-beta.6",
+    "@babel/runtime": "^7.1.2",
```
sohkai: Not sure if `aragon-client` and `aragon-messenger` need the `@babel/runtime` library, but their old babel configs were strangely using the `transform-runtime` plugin without including this dependency.

bpierre: Yes I saw the same in `aragon/aragon`, it was probably a dependency of another module.

sohkai:
@0x6431346e I've had to upgrade the `ava` dependency to add support for `babel@7`. [It looks like most of the breaking changes revolve around `t.throws()` or `t.notThrows()`](https://github.com/avajs/ava/releases), and the tests are still passing, but might be good to sanity check their release notes :).

packages/aragon-wrapper/src/ens/index.test.js
```diff
@@ -33,7 +33,7 @@ test('should resolve address for node', (t) => {
   // arrange
   const hackyOptions = {
     provider: {
-      sendAsync: undefined
+      sendAsync: 2
```
sohkai: @0x6431346e Not sure exactly what this is testing (number of calls to `sendAsync`?) but this was failing and changing it to `2` fixes it.

0x6431346e: It's supposed to test that if you call `resolve` with no `opts`, the `ENS` constructor will be called with:
```
{
  provider: {
    sendAsync: undefined
  }
}
```
and that is because `opts` defaults to `{ provider: {}}` and then, we also assign sendAsync to `provider.send` (which is `undefined`) here:
```
 // Stupid hack for ethjs-ens
  if (!opts.provider.sendAsync) {
    opts.provider.sendAsync = opts.provider.send
  }
``` 
The reason it receives `2` instead of `undefined` is because ava runs tests concurrently now, and these tests share `ethjsEnsStub`...

We can fix this by modifying both tests to `test.serial('should ...`

sohkai: @0x6431346e Actually found a better solution via [`ava`'s contexts](https://github.com/avajs/ava#test-context); see https://github.com/aragon/aragon.js/pull/168/commits/65799d72eaad1269a29506239ac1eca16724430d

chris-remus:
@sohkai Should this be moved back to the Doing column, since the checklist appears to be incomplete?

chris-remus:
@sohkai will merge.

PR #173, opened 2018-10-10
Wrapper: add address lookup proxy

sohkai:
A headache when dealing with addresses is the fact that they can be checksummed, all lowercase, or all uppercase. This makes it especially painful when doing comparisons and fetching information stored in key value maps.

This attempts to alleviate this by adding a [native `Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) that wraps the exposed `permissions` observable's values so that property accesses will check all possible variations of an address.

Fixes issues with the Permissions app in aragon/aragon, where some apps are not being fetched properly due to the checksum variations.

----------------------

As further steps, we could attempt to introduce checksum purity (we ensure that we either always or never have checksums) into internal state, start using some sort of `Address` type, etc.

### DIFF
--- a/packages/aragon-wrapper/src/utils/index.js
+++ b/packages/aragon-wrapper/src/utils/index.js
@@ -1,4 +1,5 @@
-import Proxy from '../core/proxy'
+import { isAddress, toChecksumAddress } from 'web3-utils'
+import ContractProxy from '../core/proxy'
 import { getAbi } from '../interfaces'
 
 const GAS_FUZZ_FACTOR = 1.5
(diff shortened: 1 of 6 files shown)


packages/aragon-wrapper/src/utils/index.js
```diff
@@ -11,13 +12,39 @@ export function addressesEqual (first, second) {
   return first === second
 }
 
+export function makeAddressLookupProxy (target) {
+  return new Proxy(target, {
+    get (target, property, receiver) {
```
izqui: Wouldn't it be simpler to also capture `set` and make sure all addresses are in the same format (lowercased)? Then the getter would only need to convert to lowercase and always use that as the key.

sohkai: Sounded easy on paper, but what an adventure (a5056d8).

Enforcing that the addresses are always lowercase revealed a few more address checks that weren't safe before, but also made it harder to use `dotprop` since `dotprop.get()` doesn't go through the normal getter before checking if an instance exists (https://github.com/sindresorhus/dot-prop/blob/master/index.js#L31; related issue: https://github.com/sindresorhus/dot-prop/issues/46).

sohkai:
@0x6431346e ❤️ Thanks! I didn't see any tests for the utils yet so I took the lazy approach 😄. I've just fixed them for the linter :).

chris-remus:
@izqui will have another look at this.

PR #176, opened 2018-10-10
Wrapper: Optimize app loading by throttling before doing eth_calls and caching values

izqui:
This reduces the amount `eth_call`s that aragon.js does on startup from **980** to 78, see [before](https://gist.github.com/izqui/7b838e1ce3f4de579d77a05532a00d77) and [after](https://gist.github.com/izqui/7b838e1ce3f4de579d77a05532a00d77#file-aragonjs-load-throttle).

**Edit**: Got it to [48 calls](https://gist.github.com/izqui/7b838e1ce3f4de579d77a05532a00d77#file-aragonjs-load-throttle-and-skip) with the changes introduced in 8383a37338fcad1fe7e11bdd646140b1761e8c1e

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -292,6 +292,10 @@ export default class Aragon {
       .flatMap(
         (apps) => Promise.all(
           apps.map(async (app) => {
+            if (!app.appId || !app.codeAddress) {
+              return app
+            }
+
             const cacheKey = `${app.appId}.${app.codeAddress}`
             const cachedAppInfo = dotprop.get(appInfoCache, cacheKey)
 
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -202,8 +202,15 @@ export default class Aragon {
         return [permissions, eventSet.add(event.event)]
       }, [makeAddressMapProxy({}), new Set()])
       // Skip until we have received events from all event subscriptions
+      // Note that this is safe as the ACL will always have to emit both
+      // ChangePermissionManager and SetPermission events every time a
+      // permission is created
       .skipWhile(([ permissions, eventSet ]) => eventSet.size < aclObservables.length)
       .map(([ permissions ]) => permissions)
+      // Throttle so it only continues after 30ms without new values
+      // Avoids DDOSing subscribers as during initialization there may be
+      // hundreds of events processed in a short timespan
+      .debounceTime(30)
       .publishReplay(1)
     this.permissions.connect()
   }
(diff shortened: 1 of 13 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -213,6 +224,10 @@ export default class Aragon {
       codeAddress: values[2],
       isForwarder: values[3]
     }))
+
+    this.appProxyValuesCache[proxyAddress] = appProxyValues
```
bpierre: Shall we push these in an array instead, so that we can have a max limit of cached values?

packages/aragon-wrapper/src/index.js
```diff
@@ -263,11 +291,24 @@ export default class Aragon {
...
+            const appInfo =
+              cachedAppInfo ||
+              getAragonOsInternalAppInfo(app.appId) ||
+              (await this.apm
+                .getLatestVersionForContract(app.appId, app.codeAddress)
+                // Just silence any errors
+                .catch(() => {}))
+
+            if (!cachedAppInfo && appInfo) {
+              dotprop.set(appInfoCache, cacheKey, appInfo)
```
bpierre: Isn’t `appInfoCache` going to grow indefinitely?

izqui:
Awesome @sohkai! Changes look good to me. 

However, what's the reason for the global cache? An app proxy [can only be in one DAO](https://github.com/aragon/aragon.js/blob/76688af73490aed517f54b85f8cf2592313b3941/packages/aragon-wrapper/src/index.js#L266)

packages/aragon-wrapper/src/index.js
sohkai: It will, based on how many apps (combinations of `appId` and `base contract address`) you install. I don't think we need to worry about this, as it'll likely stay very low (on current DAOs this will be 5-6), but we can start thinking about a cache invalidation strategy once it's a problem.

sohkai:
Choice for the global cache was to hide it from API consumers. I don't envision this getting huge, but if it does, because someone's switched between 1000 DAOs without refreshing, we can think about a cache invalidation strategy (e.g. restart it on every DAO?).

izqui:
Ok. I wouldn't worry about invalidation for the time being.

packages/aragon-wrapper/src/index.js
bpierre: Yes I don’t think it would be a problem right now, but I was more thinking about the app running for a very long time, browsing several DAOs (e.g. someone having a pinned tab in the browser only for Aragon). A strategy could be to use a fixed-size array instead of an object.

sohkai: Made https://github.com/aragon/aragon.js/issues/181 for this, let's worry about it later :).

PR #177, opened 2018-10-10
Allow specifying tokenName and tokenSymbol in newDAO

### LINKED ISSUE #102: wrapper/template: accept and pass the token name and symbol to newToken()
`newToken()` [now accepts](https://github.com/aragon/aragon-apps/commit/babd22b0e9c34de9dcc4be9fcb1e76e9609fc05d#diff-2bcce1990117cf15e10def37ec15ebd3) a token name and symbol for both the Democracy and Multisig templates.

https://github.com/aragon/aragon.js/blob/efa6490c486949350df844cff1c09e73c3359a2a/packages/aragon-wrapper/src/templates/index.js#L71

perissology:
fixes #102

I added `tokenName` & `tokenSymbol` as params to `newDAO` b/c it looks like `aragon/aragon` was collecting that information, but it wasn't being used here.

### DIFF
--- a/packages/aragon-wrapper/src/templates/index.js
+++ b/packages/aragon-wrapper/src/templates/index.js
@@ -33,8 +33,8 @@ const templates = {
 
 const Templates = (web3, apm, from) => {
   const minGasPrice = web3.utils.toWei('20', 'gwei')
-  const newToken = async (template, name) => {
-    const call = template.methods.newToken(name, name)
+  const newToken = async (template, name, symbol) => {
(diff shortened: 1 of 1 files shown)


sohkai:
@perissology I've pushed a commit to encapsulate the parameters for each transaction so we can easily add additional options (e.g. gas price and gas overrides) to each of them.

I've also removed the default gas, as we've pushed rinkeby's gas limit so much that wallets' gas estimation don't cause problems with new instances now.

PR #178, opened 2018-10-12
add kernel to apps stream

### LINKED ISSUE #156: Have a way to access the Kernel ABI
Potentially we could include the `kernel` as part of the apps list, or in a separate interface.

perissology:
resolves #156 

The kernel was being added to the apps stream in `aragon/core`. Moved to `aragon/wrapper` & the abi is now available.

We may want to call `makeProxy(proxyAddress, 'KernelProxy', this.web3, this.kernelProxy.initializationBlock)` in the `getAppProxyValues` function. This would allow us to fetch set the `appId` correctly as well as `codeAddress`. I didn't do this b/c `aragon/core` wasn't using those properties, and the artifacts & abi for `KernelProxy` was not included (wasn't exactly sure where those are generated from aragonOS?)


perissology:
> It actually doesn't have an appId... so I think just 'kernel' is fine :).

Isn't it in [KernelConstants.sol](https://github.com/aragon/aragonOS/blob/dev/contracts/kernel/KernelConstants.sol#L21)?

sohkai:
@perissology Ah yes, `KERNEL_APP_ID` is _technically_ the `appId`, but the `KernelProxy` itself doesn't have an `appId` storage slot so we wouldn't be able to use `getAppProxyValues()` as it is.

We could detect the proxy address given in `getAppProxyValues()` is the `KernelProxy`.

sohkai:
@perissology merged with master and added a commit that fetches the Kernel's app id from its contract.

packages/aragon-wrapper/src/index.js
```diff
@@ -222,39 +222,61 @@ export default class Aragon {
...
+      proxyValues = await Promise.all([
+        // Use Kernel ABI
+        this.kernelProxy.call('KERNEL_APP_ID').catch(() => null),
+        // Use ERC897 proxy ABI
+        kernelProxy
+          .call('implementation')
+          .catch(() => kernelProxy
+            // Fallback to old non-ERC897 proxy implementation
+            .call('getCode')
+            .catch(() => null)
```
sohkai: Note that, this bit doesn't work at all for 0.5 organizations built on aragonOS 3.1.0, as `KernelProxy` didn't expose a standard way of getting its backing `Kernel` contract.

It now implements ERC897, so `implementation()` will fetch its `codeAddress`.

packages/aragon-wrapper/src/index.js
```diff
@@ -222,39 +222,61 @@ export default class Aragon {
...
-      appId: values[1],
-      codeAddress: values[2],
-      isForwarder: values[3]
-    }))
-
-    appProxyValuesCache[proxyAddress] = appProxyValues
+    let proxyValues
+
+    if (addressesEqual(proxyAddress, this.kernelProxy.address)) {
+      const kernelProxy = makeProxy(proxyAddress, 'AppProxy', this.web3, this.kernelProxy.initializationBlock)
```
izqui: Is there any reason for using the `AppProxy` interface for the kernel proxy?

packages/aragon-wrapper/src/index.js
```diff
@@ -278,17 +300,19 @@ export default class Aragon {
...
+      })
+      // Get proxy values
       .switchMap(
         (appAddresses) => Promise.all(
-          appAddresses.map((app) => this.getAppProxyValues(app))
+          appAddresses.map((app) => this.getProxyValues(app))
         )
       )
-      .map(
-        (appMetadata) => appMetadata.filter((app) => this.isApp(app))
```
izqui: Is the `isApp` filter removed? Can't find where it is being checked right now.

This check is quite important, otherwise apps from other DAOs that just happen to be mentioned in the ACL will show up as DAO apps.

packages/aragon-wrapper/src/index.js
izqui: Is there any reason for using the `AppProxy` interface for the kernel proxy?

packages/aragon-wrapper/src/index.js
izqui: Is the `isApp` filter removed? Can't find where it is being checked right now.

This check is quite important, otherwise apps from other DAOs that just happen to be mentioned in the ACL will show up as DAO apps.

packages/aragon-wrapper/src/index.js
izqui: Shouldn't it have the `KernelProxy` interface?

izqui: Shouldn't it have the `KernelProxy` interface?

izqui: Shouldn't it have the `KernelProxy` interface?

perissology: That is code that @sohkai added, but I'm pretty sure it was b/c there is no `KernelProxy` abi in `aragon-wrapper/abi/aragon` right now

perissology:
> @perissology merged with master and added a commit that fetches the Kernel's app id from its contract.

This doesn't seem to be working for me. It is returning the wrong `appId`:

`appId: "0x01681895e4ff11b4cff40d493bc869e47e808d31e2f8c026bcda73c8e2320d8c"`

however it should be `0x3b4bf6bf3ad5000ecf0f989d5befde585c6860fea3e574a4fab4c49d1c177d9c`

was it working for you?

sohkai:
@perissology It is most likely because you're using it on a 0.5 org. [At that point, on `aragonOS@3.1.1`, the kernel's app id was calculately wrongly.](https://github.com/aragon/aragonOS/blob/4783065ecde8c648c712ce8dc0b36a0e409fdf42/contracts/kernel/KernelStorage.sol#L9)

packages/aragon-wrapper/src/index.js
sohkai: We could add `KernelProxy`, or just a generic `ERCProxy` abi instead. I mainly used `AppProxy` out of laziness (we only needed it for `implementation`, which is available in `AppProxy`).

PR #182, opened 2018-10-16
add external tx intents

### LINKED ISSUE #120: Add ability to handle external transaction intents
It would be interesting to let a frontend create transactions for external contracts (and not just the app's listed contract) in cases where an app needs to communicate to multiple contracts (e.g. in patterns where the initial app contract is a factory for a few other contracts).

The built transaction should probably still go through the `transactions` observable, and any frontends should display them with slightly different information (e.g. "This will send a transaction to an external contract at `0x...`")

ewingrj:
resolves #120 

note: this code will not correctly handle overloaded functions. It appears the current implementation of AragonApp will not handle overloaded functions for either the app, or an external contract. I'll open an issue for that


sohkai:
@perissology It's look good, but I think we'll take a closer look at this next week. Originally I thought #120 would be required for token approvals but we've since tried to automate it in #167.

We'll also have to support the frontend for it in aragon/aragon (adjust the signing panel to show that it's an transaction to an external contract that could be potentially dangerous).

sohkai:
@perissology Sorry for the long hold up on this. Don't know when we'll be able to prioritize this, since we need to make adjustments to aragon/aragon to show external intents (issue: https://github.com/aragon/aragon/issues/536).

sohkai:
Superseded by #328.

PR #188, opened 2018-10-23
Add support v0.6 kits

### LINKED ISSUE #185: Update to new dao-kit ABIs

izqui:
No longer include ABIs from NPM, but load them dynamically from the kits' APM repos.

~It seems that we made some changes to how the client uses the templates API, as it was incompatible.~

Closes #185.

### DIFF
--- a/packages/aragon-wrapper/src/templates/index.js
+++ b/packages/aragon-wrapper/src/templates/index.js
@@ -1,34 +1,17 @@
-import { templates as templateArtifacts } from '@aragon/templates-beta'
 import { toWei } from 'web3-utils'
 import { resolve as ensResolve } from '../ens'
 
 const zeroAddress = '0x0000000000000000000000000000000000000000'
 
-// TODO: Load template info dynamically from APM content package.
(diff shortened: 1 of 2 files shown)


packages/aragon-wrapper/src/templates/index.js
```diff
@@ -76,16 +50,13 @@ const Templates = (web3, apm, from) => {
...
       if (!contractAddress) throw new Error('No template contract found for that appId')
 
-      const template = new web3.eth.Contract(
-        tmplObj.abi,
-        contractAddress
-      )
+      const template = new web3.eth.Contract(abi, contractAddress)
 
-      const token = await newToken(template, tokenParams)
+      const token = await newToken(template, [tokenParams, tokenParams])
```
izqui: Hack until the client properly sends the token name and token symbol rather than just the organization name as `tokenParams`

sohkai: I think we can remove this now?

izqui:
@sohkai I just cleaned up the branch to only have the needed changes!

izqui:
Changes looking good @sohkai!

PR #189, opened 2018-10-23
wrapper: avoid using 20gwei by default for new dao transactions

sohkai:
Let the ethereum provider decide how much to price the new dao transactions.


sohkai:
Already pushed to master.

PR #195, opened 2018-10-24
wrapper: upgrade radspec

sohkai:
Upgrades radspec to the fixed build version, and removes `date-fns` since we only used it for a trivial `isBefore()` check.


sohkai:
@izqui I've replaced `date-fns` with `date1.getTime() < date2.getTime()` since [that's what they're doing](te-fns/blob/master/src/isBefore/index.js#L31).

PR #199, opened 2018-10-26
Fix apmOptions in Template

izqui:
**This is a breaking change in the API**


packages/aragon-wrapper/src/index.js
```diff
@@ -88,22 +125,17 @@ const appInfoCache = {}
 export default class Aragon {
   constructor (daoAddress, options = {}) {
     const defaultOptions = {
-      provider: detectProvider(),
       apm: {},
-      defaultGasPriceFn: () => {
-        return toWei('20', 'gwei')
```
sohkai: I've opted just to remove this default, since ethereum providers are getting pretty good at handling this by themselves.

PR #200, opened 2018-10-26
Implement radspec postprocessing for humanizing addresses and roles

izqui:
This is sort of a temporary hack as this should probably be built as an extension to radspec in some way, but I'd like to have a bit more of time to think about how that API would look like. Also in the case of addresses we will probably want to do this directly in the client so we can show the entity component.

I'm having doubts about putting the replacements in quotes as you can see in the screenshot, do you think it would look better if we made the text bold @jounih?

Before:
2018-10-26 at 11 37 20 am

After:
2018-10-26 at 11 53 52 am

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -888,6 +888,7 @@ export default class Aragon {
       if (!expression) return step
 
       let description
+      let annotatedDescription
       try {
         description = await radspec.evaluate(
           expression,
@@ -899,12 +900,15 @@ export default class Aragon {
         )
 
         try {
-          description = await this.postprocessRadspecDescription(description)
+          const processed = await this.postprocessRadspecDescription(description)
+          description = processed.description
+          annotatedDescription = processed.annotatedDescription
         } catch (err) { }
       } catch (err) { }
 
       return Object.assign(step, {
         description,
+        annotatedDescription,
         name: app.name,
         identifier: app.identifier
       })
@@ -915,7 +919,7 @@ export default class Aragon {
    * Look for known addresses and roles in a radspec description and substitute them with a human string
    *
    * @param  {string} description
-   * @return {Promise<string>} modified description
+   * @return {Promise<Object>} description and annotated description
    */
   async postprocessRadspecDescription (description) {
     const addressRegex = /\b0x[a-fA-F0-9]{40}\b/g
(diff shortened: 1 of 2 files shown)


jounih:
Yes it would be more readable to have those in bold I think

izqui:
Good idea @bpierre, added 693bd49 with an implementation of `annotatedDescription`. It is added to the returned object in `describeTransactionPath`.

For this description: 

```
Grants 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF the ability to perform actions of role 0x0b719b33c83b8e5d300c521cb8b54ae9bd933996a14bef8c2f4e0285d2d2400a on 0x2a9446C6721fE12dFACA23b8e86fcFE2C778a7b3
```

It returns this annotated description:

```
[
  { "type": "text", "value": "Grants" },
  { "type": "any-account" },
  { "type": "text", "value": "the ability to perform actions of role" },
  { "type": "role", "role": {
      "name": "Create permissions",
      "id": "CREATE_PERMISSIONS_ROLE",
      "bytes": "0x0b719b33c83b8e5d300c521cb8b54ae9bd933996a14bef8c2f4e0285d2d2400a"
    }
  },
  { "type": "text", "value": "on" },
  { "type": "app", "app": {...} } // App object
]
```

And this processed description:
```
Grants 'Any account' the ability to perform actions of role 'Create permissions' on 'ACL'
```

sohkai:
@bpierre the last commit changed the API of the annotated descriptions, such that in the above example (`Grants 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF the ability to perform actions of role 0x0b719b33c83b8e5d300c521cb8b54ae9bd933996a14bef8c2f4e0285d2d2400a on 0x2a9446C6721fE12dFACA23b8e86fcFE2C778a7b3`), it would give you the following:

```js
[
  { "type": "text", "value": "Grants" },
  { "type": "any-account", "value": "0xfff..." },
  { "type": "text", "value": "the ability to perform actions of role" },
  { "type": "role", "value": {
      "name": "Create permissions",
      "id": "CREATE_PERMISSIONS_ROLE",
      "bytes": "0x0b719b33c83b8e5d300c521cb8b54ae9bd933996a14bef8c2f4e0285d2d2400a"
    }
  },
  { "type": "text", "value": "on" },
  { "type": "app", "value": {...} } // App object
]
```

Every annotation node is just "type" and "value" rather than a different key based on the type.

izqui:
Tested and working well. Nice improvement!

PR #202, opened 2018-10-27
wrapper: refactor radspec annotations

sohkai:
- Unifies the annotation format across all the different types (text, addresses, bytes, apps, roles) so that they're all in the format of `{ type, value }`
- Adds two types (generic addresses and bytes)
- Simplifies the parsing implementation

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -898,13 +898,13 @@ export default class Aragon {
           },
           { ethNode: this.web3.currentProvider }
         )
-
-        try {
-          const processed = await this.postprocessRadspecDescription(description)
-          description = processed.description
-          annotatedDescription = processed.annotatedDescription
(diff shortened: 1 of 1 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -898,14 +898,14 @@ export default class Aragon {
           },
           { ethNode: this.web3.currentProvider }
         )
-
-        try {
-          const processed = await this.postprocessRadspecDescription(description)
-          description = processed.description
-          annotatedDescription = processed.annotatedDescription
-        } catch (err) { console.log(err) }
```
sohkai: Wasn't sure if there was a particular exception being caught here; as far as I can tell, `postprocessRadspecDescription()` shouldn't throw if you provide it a string

izqui: You are right, forgot to remove it after testing

PR #205, opened 2018-10-31
Depend on @aragon/os for ABIs

### LINKED ISSUE #46: Depend on @aragon/os for ABIs rather than having them hardcoded in node-aragon
Requires https://github.com/aragon/aragonOS/issues/213

We have an ABIs directory that is updated manually with the latest version of the aOS ABIs.

galactusss:
Solve #46


sohkai:
Looking good, let's wait for #187 before merging this since this will also cause incompatibilities.

I've also added a commit since we export objects with the form:

```json
{
  "abi" : {
    ...
  }
}
```

So you need the `.abi` :).

Issue #133, opened 2018-08-23
Migrate to IndexDB / FileSystem API?


sohkai:
[Good roundup of storage limits](https://www.html5rocks.com/en/tutorials/offline/quota-research/#toc-overview).

aragon.js-wrapper could easily move to [localForage](https://github.com/localForage/localForage) instead of lowdb and leverage IndexDB as a starting point, although we'd have to think about node support.

The main problem with localStorage is the low limit (5-10mb, ~5-10 million characters) on most browsers. Currently, syncing `ngo.aragonid.eth` on Rinkeby is about ~19k chars, but this is a DAO with very little information.

PR #208, opened 2018-11-12
Migrate to IndexDB using localForage

galactusss:
Solve part of #133


packages/aragon-wrapper/package.json
```diff
@@ -73,6 +73,7 @@
     "ethjs-ens": "^2.0.1",
     "homedir": "^0.6.0",
     "js-sha3": "^0.7.0",
+    "localforage": "^1.7.3",
     "lowdb": "^1.0.0",
```
sohkai: We could likely get rid of the `lowdb` and slim down our `lodash` requirements a lot!

packages/aragon-wrapper/src/cache/index.js
```diff
@@ -23,44 +19,36 @@ export default class Cache {
...
-    // Set default cache state
-    this.db.defaults({}).write()
 
     // Set up the changes observable
     this.changes = new Subject()
+
+    // Set up cache DB
+    this.db = localforage.createInstance({
+      driver: [localforage.INDEXEDDB, localforage.LOCALSTORAGE],
+      name: `localforage/${this.prefix}`
```
sohkai: It'd be really nice if the localstorage set up was the same as `lowdb`, but I imagine the chances of this being possible are quite slim.

packages/aragon-wrapper/package.json
galactusss: Cool, thanks for the comments. I'll change that and a few things Daniel pointed me

packages/aragon-wrapper/src/cache/index.js
galactusss: I'm not entirely sure I understand, localforage decide to use one of the drivers depending order. You meant that in case localstorage is selected, use the same config as lowdb?
I can dig deeper about that

packages/aragon-wrapper/src/cache/index.js
```diff
@@ -23,44 +19,36 @@ export default class Cache {
...
   getCacheKeyPath (key) {
     return `${this.prefix}.${key}`
   }
 
-  async set (key, value) {
-    // Some lowdb adapters are synchronous while others are asynchronous so
-    // let's always wrap it in a promise
-    await Promise.resolve(this.db.set(
+  set (key, value) {
+    this.db.setItem(
```
AquiGorka: I'm not familiar with this part of the codebase so maybe you could help me understand here:
Why are you not `await`ing for the operation to finish (like it was before)? [`setItem`](https://github.com/localForage/localForage#callbacks-vs-promises) returns a `promise`:

```js
set (key, value) {
  await this.db.setItem(
    this.getCacheKeyPath(key),
    value
  )
  this.changes.next({ key, value })
}
```

galactusss: You are right, I need to use await the operation. Daniel point the same thing to me the other day. Thank you

packages/aragon-wrapper/src/cache/index.js
sohkai: Yeah, if localstorage is used, it'd be nice if the data schema was the same (a quick a/b test between the two would show the differences). Again, I think the chances of this being possible is _really_ low and not a point to stress too much on.

sohkai:
@galactusss is this something we should review again? I noticed there was still a merge conflict that was commited in the package.json.

galactusss:
@sohkai I fix the merge conflicts. I missed the bump of radspec 👍

galactusss:
I closed and reopened to run travis again

packages/aragon-wrapper/src/cache/index.js
```diff
@@ -10,57 +8,35 @@ export default class Cache {
...
     this.changes.next({ key, value })
   }
 
-  get (key, defaultValue) {
-    return this.db.get(
-      this.getCacheKeyPath(key)
-    ).value() || defaultValue
-  }
-
-  update (key, transition) {
```
0x6431346e: I can't see this being used anywhere, I think it's safe to say this is not a breaking change.
What do you think @sohkai ?

sohkai: Sounds good to me! The breaking change won't be a problem for now anyway as we're still on beta versions :).

sohkai:
For the `observe()` issue: we could use [`concat`](https://rxjs-dev.firebaseapp.com/api/index/function/concat) to merge the promise from `get()` into the changes observable:

```js
const changes = this.changes.pipe(
  filter((change) => change.key === key),
  pluck('value')
)
return concat(this.get(key, defaultValue), changes)
```

`concat` takes in an `ObservableInput`, which can be a Promise, so I think this should just work.I'm a little bit worried about a potential race condition between the `this.get()` and `this.changes()` (where one or more new `set` operations may update the value before `get` finishes), which may cause the starting value to be in the middle of the changes stream, but I think this shouldn't be an issue if we wrap `this.changes` with a defer:

```js
const changes = defer(() =>
  this.changes.pipe(
    filter((change) => change.key === key),
    pluck('value')
))
return concat(this.get(key, defaultValue), changes)
```

galactusss:
@0x6431346e Thank you for adding the fix 🎉

0x6431346e:
Looks like the build is `7.4 KB` heavier now :disappointed: ...getting rid of lowdb did not help.

sohkai:
@0x6431346e likely because we're still using lodash somewhere? We could probably slim that down, since `lowdb` is just a really thing wrapper around lodash.

packages/aragon-wrapper/src/cache/index.test.js
```diff
@@ -10,33 +10,57 @@ test.afterEach.always(() => {
...
+  observable.subscribe(value => {
+    emissionNumber++
+    // first value should be 3 (the default) because getItem returns falsy
+    if (emissionNumber === 1) t.is(value, 1)
+    if (emissionNumber === 2) t.is(value, 10)
+    if (emissionNumber === 3) t.is(value, 11)
+    if (emissionNumber === 4) t.is(value, 12)
+  })
+
+  // these will be ignored because they happen before `get` finishes
```
sohkai: ❤️ Thanks for testing this!!

0x6431346e:
> @0x6431346e likely because we're still using lodash somewhere? We could probably slim that down, since `lowdb` is just a really thing wrapper around lodash.

Can't see any reference of lodash in our code.. :thinking: 

image

What I can tell from this though is that our dependencies use a lot of different versions of lodash/underscore.

PR #209, opened 2018-11-13
Add description postprocessing for APM app names and Kernel namespaces

izqui:


packages/aragon-wrapper/src/core/aragonOS/index.js
```diff
@@ -17,6 +17,12 @@ const APP_MAPPINGS = {
   [oldWrongAppId('evmreg')]: 'EVM Script Registry'
 }
 
+const KERNEL_NAMESPACES = {
```
AquiGorka: Would you consider using a Map here:
```js
const KERNEL_NAMESPACES = new Map([
  [soliditySha3('core'), 'Core'],
  [soliditySha3('app'), 'Default apps'],
  [soliditySha3('base'), 'App code'],
])

function getKernelNamespace (hash) {
  if (KERNEL_NAMESPACES.has(hash)) {
    return { name: KERNEL_NAMESPACES.get(hash), hash }
  }
}
```

sohkai: 👍 we could update `APP_MAPPINGS` to be a map too.

0x6431346e: Updated in 503a1c8 , I will merge this as I have some tests on another branch `postprocessRadspecDescription`

PR #212, opened 2018-11-14
feat: default external events' fromBlock to app's initialzation block

sohkai:
Fixes https://github.com/aragon/aragon-apps/issues/569

Not really the nicest, but this particular bit of async / sync in `runApp()` was a deep, deep rabbit hole of race conditions. In particular, it was a constant battle to make sure event handlers were attached synchronously for webworkers, and, in the case of iframes, delaying our handshake message until we were sure everything was hooked up.

As such, `runApp()` was split into two parts, to allow an async "preparation" phase for collecting necessary state (i.e. the app's initialization block). The first phase returns a function that an implementation should later invoke with a message provider to actually "connect" a running app to aragon.js's message handlers. See [related PR in `aragon/aragon`](https://github.com/aragon/aragon/pull/476).

Another thing to note is that `runApp()` now ignores any updates to the apps observable. Updating running apps in the past would've totally broken them (their entire context would've been reset), so this makes it safer, but does leave open the question of how we should support live updates to apps (probably another observable containing update requests that we `switchmap` on?).


packages/aragon-wrapper/src/index.js
```diff
@@ -590,66 +590,77 @@ export default class Aragon {
...
    *         The address of the app proxy.
-   * @return {Object}
+   * @return {Promise<function>}
    */
-  runApp (sandboxMessengerProvider, proxyAddress) {
-    // Set up messenger
-    const messenger = new Messenger(
-      sandboxMessengerProvider
-    )
+  async runApp (proxyAddress, connect) {
```
0x6431346e: The `connect` parameter doesn't seem to be used

0x6431346e: The connect parameter doesn't seem to be used

packages/aragon-client/src/index.js
```diff
@@ -91,17 +91,21 @@ export class AppProxy {
...
     const contract = {
-      events: (fromBlock = 0) => {
+      events: (fromBlock) => {
+        const eventArgs = [
+          address,
+          jsonInterface.filter(
+            (item) => item.type === 'event'
+          )
+        ]
+        if (fromBlock != null) {
```
0x6431346e: Do we want to change this to `fromBlock !== null && fromBlock !== undefined` ? :man_shrugging:

packages/aragon-wrapper/src/index.js
```diff
@@ -590,66 +590,77 @@ export default class Aragon {
...
+    // Finish the observable, so our running contexts don't get reinitialized if new apps appear
     // NOTE: we **CANNOT** use this.apps here, as it'll trigger an endless spiral of infinite streams
-    const proxy$ = this.appsWithoutIdentifiers
-      .map((apps) => apps.find(
-        (app) => addressesEqual(app.proxyAddress, proxyAddress))
-      )
-      // TODO: handle undefined (no proxy found), otherwise when calling app.proxyAddress next, it will throw
-      .map(
-        (app) => makeProxyFromABI(app.proxyAddress, app.abi, this.web3, this.kernelProxy.initializationBlock)
+    const appProxy = await this.appsWithoutIdentifiers
```
sohkai: Realized with the `.take(1)` we could probably just use `this.apps` as well now

packages/aragon-client/src/index.js
sohkai: Hmm, actually thinking about doing a `typeof` check instead, since it should only be a number.

PR #230, opened 2019-01-03
[WIP] Simpler API using events and promises

### LINKED ISSUE #226: Return promises instead of observables
@0x6431346e commented on [Fri Dec 14 2018](https://github.com/aragon/aragon-cli/issues/300)

See https://github.com/aragon/aragon-app-dx/issues/2

Would be nice to add an option when initializing the app and then have promises returned.
```
const app = new AragonApp({ withPromises: true })

app.increment(1).then(
  (txHash) => console.log(`Success! Incremented in tx ${txHash}`),
  (err) => console.log(`Could not increment: ${err}`)
)
```
instead of
```
const app = new AragonApp()

app.increment(1).subscribe(
  (txHash) => console.log(`Success! Incremented in tx ${txHash}`),
  (err) => console.log(`Could not increment: ${err}`)
)
```

0x6431346e:
Summary:
- describeScript (emits **once**) :heavy_check_mark: 
- web3Eth (emits **once**) :heavy_check_mark: 
- call (emits **once**) :heavy_check_mark:  (now as `contract.call`)
- intents (emits **once**) :heavy_check_mark:  (now as `contract.intent`)
- events (emits **updates**) :heavy_check_mark: (now as `contract.on('event')`)
- external 
  - call (emits **once**) :heavy_check_mark: (now as `externalContract(...).call`)
  - events (emits **updates**) :heavy_check_mark: (now as `externalContract(...).on('event')`)
- accounts (emits **updates**) :heavy_check_mark: (now as `accounts.on('update')`)
- network (emits **updates**) :heavy_check_mark: (now as `network.on('update')`)
- state (emits **updates**) :heavy_check_mark: (now as `state.on('update')`)
- context (emits **updates**) :x: removed
- store :heavy_check_mark: no longer returns nor takes the events param.
- dispatch :heavy_check_mark: new method to feed other events: `externalContract(address, abi).on('event', dispatch)`
- ~identify~ (sync) 
- ~notify~ (sync)
- ~cache~ (sync)

TODO:
- [ ] add tests
- [ ] update docs
- [ ] hide providers from the users

Closes #226

### DIFF
--- a/packages/aragon-client/src/index.js
+++ b/packages/aragon-client/src/index.js
@@ -1,35 +1,104 @@
 import Messenger, { providers } from '@aragon/messenger'
-import { defer } from 'rxjs/observable/defer'
-import { empty } from 'rxjs/observable/empty'
 import { fromPromise } from 'rxjs/observable/fromPromise'
 import { merge } from 'rxjs/observable/merge'
 import EventEmitter from 'events'
-import { ReplaySubject } from 'rxjs/Rx'
+import { ReplaySubject, Subject } from 'rxjs/Rx'
 
-/**
- * @private
- */
-export const AppProxyHandler = {
-  get (target, name, receiver) {
-    if (name in target) {
-      return target[name]
-    }
+export class ContractAPI extends EventEmitter {
+  constructor(rpc) {
+    super()
+    this.rpc = rpc
+    // lazily initialize the events (after the first subscribe)
+    this.once('newListener', (event, listener) => {
+      if (event === 'event') {
+        // this event (newListener) is fired BEFORE the listener has been added
+        // defer this with setTimeout to initialize the events after the listender was added 
+        setTimeout(() => this.initializeEvents(), 0)
+      }
+    })
+  }
 
-    return function (...params) {
-      return target.rpc.sendAndObserveResponse(
-        'intent',
-        [name, ...params]
-      ).pluck('result')
-    }
+  initializeEvents () {
+    this.rpc.sendAndObserveResponses('events')
+      .pluck('result')
+      .subscribe(value => {
+        this.emit('event', value)
+      })
+  }
+
+  intent (name, ...params) {
+    return this.rpc.sendAndObserveResponse(
+      'intent',
+      [name, ...params]
+    )
+      .pluck('result')
+      .toPromise()
+  }
+
+  call (method, ...params) {
+    return this.rpc.sendAndObserveResponse(
+      'call',
+      [method, ...params]
+    )
+      .pluck('result')
+      .toPromise()
   }
 }
 
-export class NetworkAPI extends EventEmitter {
-  constructor (rpc) {
+export class ExternalContractAPI extends EventEmitter {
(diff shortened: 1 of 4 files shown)


sohkai:
Looking really good!

Something I didn't pick up in https://github.com/aragon/aragon-app-dx/issues/2 was the API surrounding `app.external()` and `state()`. We should be careful to make sure it's still easy to provide additional observables / event emitters to `createAppState()` so that it can combine multiple event streams, like we do in most apps: https://github.com/aragon/aragon-apps/blob/master/apps/finance/app/src/script.js#L152

packages/aragon-client/src/index.js
```diff
@@ -19,12 +21,29 @@ export const AppProxyHandler = {
...
+    this.rpc = rpc
+    this.observable = new ReplaySubject(1)
+    this.rpc.sendAndObserveResponses('network')
+      .pluck('result')
+      .do(value => { this.emit('update', value) })
+      .subscribe(this.observable)
+  }
+
+  get () {
+    return this.observable.toPromise()
```
sohkai: Not 100% sure what the behaviour is with a `ReplaySubject()`, but I think we still need to do a `.first()` to end the observable before calling `toPromise()`?

luisivan:
Oh my god, this is gold. Honestly the only reason why I'm cannot create the Aragon apps I want to create is because I don't have the time or hustle to interact with observables and RxJS.
What's blocking this @0x6431346e? Reviews?

0x6431346e:
> Oh my god, this is gold. Honestly the only reason why I'm cannot create the Aragon apps I want to create is because I don't have the time or hustle to interact with observables and RxJS.
> What's blocking this @0x6431346e? Reviews?

Not necessarily, it's missing tests and considering this was meant to replace the existing API, a lot of them. Also since it's a big change that will affect all the devs the more feedback the better (especially from devs that still want to use rxjs - to see if they can seamlessly migrate).

That being said, since we'll have `api-react` soon, maybe it makes more sense for this to be a separate package as well, something like `@aragon/api-node` ? This way we also don't need to test so much and just release it as beta. Later on we can decide how much testing and support we should allocate based on the feedback from early adopters.

I'm thinking that the aragonCLI could be a consumer of this api, since we don't take advantage of rxjs at all. 🤔

sohkai:
@0x6431346e changing this to be `@aragon/api-node` sounds very reasonable, especially with the event emitter interface.

One thing I've been wanting to do is release `@aragon/api@v2.0.0` with the single emission observables modified to be just promises. I think that should be fairly transparent for both `api-react` and `api-node`, but will require a major version release since it's a breaking change.

packages/aragon-client/src/index.js
sohkai: Naw, just that `this.observable.toPromise()`'s promise won't resolve until `this.observable` itself ends, so we need to use `.first()` to end it.

sohkai:
@0x6431346e @0xGabi Since this has gone so stale, let's close it for now and come back to it in the future if we need to.

PR #232, opened 2019-01-08
Simplify the initAcl logic

2color:
## What
**Non-breaking change** 👐

1. A small refactor of the `initAcl` method. 
2. Add default value to `makeAddressMapProxy` as it's an empty object in most use cases anyways

## Why 
The refactor of `initAcl` was done in preparation for Acl caching. 

## How 
Retrieve all events instead of creating two event streams for each event type.

Because the Acl contract only emits the two events currently ([of the three defined](https://github.com/aragon/aragonOS/blob/dev/contracts/acl/ACL.sol#L57-L59)) this doesn't introduce any network overhead for the extra data.

Because the events of the two types we're interested in are now ordered in a single stream, we no longer need:

 - the added `skipWhile` operator
 - `eventSet` which was passed along to know when the first event stream is complete

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -214,20 +214,11 @@ export default class Aragon {
     const SET_PERMISSION_EVENT = 'SetPermission'
     const CHANGE_PERMISSION_MANAGER_EVENT = 'ChangePermissionManager'
 
-    const aclObservables = [
-      SET_PERMISSION_EVENT,
-      CHANGE_PERMISSION_MANAGER_EVENT
-    ].map(event =>
-      this.aclProxy.events(event)
-    )
-
-    // Set up permissions observable
-
     // Permissions Object:
-    // app -> role -> { manager, allowedEntities -> [ entities with permission ] }
-    this.permissions = Observable.merge(...aclObservables)
-      // Keep track of all the types of events that have been processed
-      .scan(([ permissions, eventSet ], event) => {
+    // { app -> role -> { manager, allowedEntities -> [ entities with permission ] } }
+    this.permissions = this.aclProxy.events()
+      // Keep track of all events types that have been processed
+      .scan((permissions, event) => {
         const eventData = event.returnValues
 
         // NOTE: dotprop.get() doesn't work through proxies, so we manually access permissions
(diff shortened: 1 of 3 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -237,30 +228,25 @@ export default class Aragon {
...
         }
 
         permissions[eventData.app] = appPermissions
-        return [permissions, eventSet.add(event.event)]
-      }, [makeAddressMapProxy({}), new Set()])
-      // Skip until we have received events from all event subscriptions
-      // Note that this is safe as the ACL will always have to emit both
-      // ChangePermissionManager and SetPermission events every time a
-      // permission is created
-      .skipWhile(([ permissions, eventSet ]) => eventSet.size < aclObservables.length)
```
sohkai: I'm a little bit worried about this leaking intermediate state. Not that the old implementation did it well (it only throttled one event, but emitted events for each pass of the second).

I just found out about [`.getPastEvents()`](https://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#getpastevents), which we could start using in all places to avoid producing intermediate state (given that the `.events()` just sends a `eth_getLogs` and a `eth_subscription` when using a past `fromBlock`).

This will also give us information about whether we're waiting for old logs or already up to date and waiting for new information.

My thoughts on doing this:

- Get the current block (for safety reasons below)
- Define the particular data transforms outside of the data stream, taking in a single observable and producing a single observable as return
- Create an observable that emits a single value, which streams the result of `.getPastEvents()` (with the current block) into the data transform
  - Somehow mark that we've finished loading past events, either with a flag in the result or some other mechanism
- Create an observable from `.events()`, using the current block as the `fromBlock`, that also streams into the data transform
- Combine the above observables using [`concat()`](https://rxjs-dev.firebaseapp.com/api/index/function/concat), putting the `getPastEvents()` observable first

This way, we'll know if we've dealt with past data (both to avoid intermediate state and throttle emits to later observables) and are now ready to stream new events as they come from the chain.

Safety note: it's important to use the same current block when creating both observables, as if they're done separately, there's a race condition with potentially missing some blocks (or duplicating the latest) due to when the block gets created, network conditions, etc. We might need to use `fromBlock: currentBlock + 1` for `.events()` (hopefully this also avoids another call to `eth_getLogs`), but we'd have to check the web3.js source.

packages/aragon-wrapper/src/utils/index.js
```diff
@@ -19,7 +19,8 @@ export function includesAddress (arr, address) {
   return arr.some(a => addressesEqual(a, address))
 }
 
-export function makeAddressMapProxy (target) {
+// Address map that prevents upper/lower case mistakes
```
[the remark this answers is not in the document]
2color: I understand. So whatever address format is passed in, it is converted to lower case so no checksums are assumed across the board?

```suggestion
// Address map that ensures consistent non-checksummed interpretations of addresses
```

packages/aragon-wrapper/src/index.js
2color: I like the idea of separating data reductions from the fetching layer. 

> I'm a little bit worried about this leaking intermediate state

I'm not well versed in all the ramifications of _leaking intermediate state._ I'm also not sure I fully understand when state is considered intermediate. 

Perhaps we should be specific about when we actually want an ACL state observable emitting the permissions object

I support the idea of separating data transformation from fetching. But I can't really comment on all the other suggestions of separating it into two separate streams for past events and for future events. 

Given the time constraints of the trial week and the ultimate goal implementing a caching mechanism, how would you approach this more pragmatically?

2color: Caching should reduce intermediate state leakage at least from the second run. But it's still difficult to envision how breaking it into two streams would help with the intermediate state problem.

2color: Another thought when looking at this. 

We already know the initialisation block. We can check what's the current head in the beginning of `initAcl`. This allows us to know how close we are during the processing of events to reaching the _current_ (a moving target) state

packages/aragon-wrapper/src/utils/index.js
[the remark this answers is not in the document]
sohkai: Yup, you got it :).

2color:
Closing in favour of #233

PR #233, opened 2019-01-09
ACL caching

2color:
## What

- Simplify initAcl
- Cache the permissions object (reduced ACL events)

## Why
- This builds on #232 
- A separate PR is for experimenting with caching while leaving the discussion about the initial refactor in the other PR
- Still contains a lot of logging for debugging 

## How 
- Add the ability to fetch events from a specific block number
- Cache blocks that we have finished processing

## Open Questions
- Is there a way to know that I've finished processing all the events of a block? 
  - Yes with `getPastLogs` because the call resolves with a single value

Currently I wait until I get the next block and only then do I assume that I've gotten all the events for the previous. 
This means that the last block's events are not cached and have to be fetched again. 

## Potential improvements

- Create unit tests (hopefully with as little test doubles as possible) for the caching functionality. This should help in turn with the next steps of abstracting cleaning and also provide another form of documentation for the logic.
- Update the blockNumber in cache to a later point if we're certain no relevant events came in later blocks - reduce load on node filtering from an earlier point every time.


packages/aragon-wrapper/src/core/proxy/index.js
```diff
@@ -12,12 +12,16 @@ export default class Proxy {
   }
 
   // TODO: Make this a hot observable
-  events (eventNames) {
+  events (eventNames, fromBlock) {
```
izqui:
```suggestion
  events (eventNames, fromBlock = this.initializationBlock) {
```

I am unsure whether this is allowed :D

packages/aragon-wrapper/src/index.js
```diff
@@ -214,20 +214,49 @@ export default class Aragon {
...
-    // Set up permissions observable
+    const ACL_CACHE_KEY = `acl-${aclAddress}`
+
+    // Check if we have cached ACL for this address
+    // Cache object for an ACL:
+    // [ { permissions, blockNumber } ]
+    const cached = await this.cache.get(ACL_CACHE_KEY, {})
+    const { permissions: cachedPermissions, blockNumber: cachedBlockNumber } = cached
+
+    let events
```
izqui:
```suggestion
    const events  = this.aclProxy.events(null, cachedPermissions && cachedBlockNumber)
```

If we don't need the log in L227 we can just do this

packages/aragon-wrapper/src/index.js
```diff
@@ -214,20 +214,49 @@ export default class Aragon {
...
+    if (cachedPermissions) {
+      console.log(`using cache from block ${cachedBlockNumber}`)
+      events = this.aclProxy.events(null, cachedBlockNumber)
+    } else {
+      events = this.aclProxy.events()
+    }
 
+    // A set for keeping track of the events' blockNumber
+    const blockNumbers = new Set([-1])
```
izqui: A comment on why we add `-1` to the set would be appreciated

packages/aragon-wrapper/src/core/proxy/index.js
2color: I wasn't sure. But it is: https://gist.github.com/2color/316ab17bcf1f6aa064de73653bc9df45

2color:
| Run          | fetch -> first event message | processing first message -> permissions ready |
|--------------|------------------------------|-----------------------------------------------|
| 1 with cache | 3175.10205078125ms           | 35.1689453125ms                               |
| 2 with cache | 3690.963134765625ms          | 34.9951171875ms                               |
| 3 with cache | 2444.296142578125ms          | 35.690185546875ms                             |
| 4 with cache | 5280.4609375ms               | 36.490234375ms                                |
| 5 no cache   | 3935.661865234375ms          | 150.660888671875ms                            |
| 6 no cache   | 3319.256591796875ms          | 131.905029296875ms                            |

I did a few tests to see where the lag originates.  

While this PR reduces the processing time by about 100ms (This is of course a correlate of the number of events which it bound to grow over time for active DAOs).

My suggestion moving forward would be to investigate whether it makes sense to let the cached permissions flow out of the function before waiting for the first events to arrive.

packages/aragon-wrapper/src/index.js
```diff
@@ -228,20 +239,42 @@ export default class Aragon {
...
+          blockNumbers.add(event.blockNumber)
+
+          // Save fully processed valid (non-zero) blocks that are at least 100 blocks old
+          if (lastBlockProcessed > 0 && lastBlockProcessed < (latestBlockNumber - REORG_SAFETY_BLOCK_AGE)) {
+            // clone the permissions so it can be saved without proxy
+            const permissionsToCache = Object.assign({}, permissions)
+            // cache optimistically without worrying if it succeeded
+            this.cache.set(ACL_CACHE_KEY, { permissions: permissionsToCache, blockNumber: lastBlockProcessed })
+          }
+        }
```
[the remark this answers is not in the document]
sohkai: With `pairwise()` it does mean we have to be careful to create new `permissions` objects in the scan on each run, otherwise `lastPermissions` and `currentPermissions` will be the same object.

We should also modify `makeAddressMapProxy()` so that it makes a new object with all lowercase keys to create the proxy with.

packages/aragon-wrapper/src/index.js
```diff
@@ -228,20 +239,42 @@ export default class Aragon {
...
+    const ACL_CACHE_KEY = `acl-${aclAddress}`
+    const REORG_SAFETY_BLOCK_AGE = 100
+
+    // Check if we have cached ACL for this address
+    // Cache object for an ACL: { permissions, blockNumber }
+    const cached = await this.cache.get(ACL_CACHE_KEY, {})
+    const { permissions: cachedPermissions, blockNumber: cachedBlockNumber } = cached
+
+    // When using cache, fetch events from the next block after cache
+    const eventsOptions = cachedPermissions && { fromBlock: (cachedBlockNumber + 1) }
```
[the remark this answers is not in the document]
2color:
```suggestion
    const eventsOptions = cachedPermissions ? { fromBlock: (cachedBlockNumber + 1) } : undefined
```

+1 for ternary operator. Thought when `cachedPermissions` is falsey we want to pass in `undefined`, so that the `initializationBlock` is correctly set as the fromBlock.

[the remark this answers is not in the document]
sohkai: I think both would work, but `undefined` is a little bit faster :).

packages/aragon-wrapper/src/index.js
```diff
@@ -251,36 +284,39 @@ export default class Aragon {
...
       debounceTime(30),
       publishReplay(1)
     )
+
+    if (cachedPermissions) {
+      const permissons = of(cachedPermissions).pipe(publish())
+      this.permissions = merge(permissons, fetchedPermissions).pipe(publishReplay(1))
+    } else {
+      this.permissions = fetchedPermissions
+    }
```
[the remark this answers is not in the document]
2color: Thanks for the suggestion. That does clean up the mess there 😉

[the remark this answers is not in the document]
2color: After trying this out it appears that
`this.permissions = concat(of(), fetchedPermissions).pipe(publishReplay(1))` 
just hangs. 

I went down a debugging path to try and find the root cause of this and I've discovered a bug in the implementation. It appears that `concat`enating in this part does not yield an observable that gets events from `fetchedPermissions` (a hot observable). 

This is very easy to miss, but simply modifying the ACL when `concat` does not trigger an event on the whole reduction pipieline.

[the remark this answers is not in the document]
2color: That's likely because we don't connect to `fetchedPermissions`.

I'll test that and see

packages/aragon-wrapper/src/index.js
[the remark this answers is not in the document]
2color: Addressed in https://github.com/aragon/aragon.js/pull/233/commits/7e68be156e2b72550c2578e592d0319fb663877c

packages/aragon-wrapper/src/cache/index.js
```diff
@@ -41,6 +41,16 @@ export default class Cache {
     }
   }
 
+  /**
+   * Get a standard cache key
+   *
+   * @param {string} proxyAddress
+   * @param {string} location
+   */
+  getCacheKey(proxyAddress, location) {
```
sohkai: Nitpick: wondering if this would fit better outside of the `Cache` class since it's not really related to any `Cache` instances?

packages/aragon-wrapper/src/index.js
```diff
@@ -311,12 +310,9 @@ export default class Aragon {
...
 
-    if (cachedPermissions) {
-      const permissons = of(cachedPermissions).pipe(publish())
-      this.permissions = merge(permissons, fetchedPermissions).pipe(publishReplay(1))
-    } else {
-      this.permissions = fetchedPermissions
-    }
+    const cachedPermissions$ = cachedPermissions ? of(cachedPermissions) : of()
+    this.permissions = concat(cachedPermissions$, fetchedPermissions$).pipe(publishReplay(1))
+    fetchedPermissions$.connect()
```
sohkai: Nitpick: I generally prefer grouping the `.connect()` as close as possible with the `publish()` / `share()` / etc to avoid forgetting it.

There are times where you would want to leave all the `.connect()`s to the end (e.g. if you had a bunch of synchronous observables that should all be subscribed to at the same time), but in this case it's not important as `fetchedPermissions$` is async.

packages/aragon-wrapper/src/utils/index.js
```diff
@@ -21,7 +21,12 @@ export function includesAddress (arr, address) {
 
 // Address map that ensures consistent non-checksummed interpretations of addresses
 export function makeAddressMapProxy (target = {}) {
-  return new Proxy(target, {
+  const targetLowerCaseKeys = {}
+  Object.keys(target).forEach(addr => {
+    targetLowerCaseKeys[addr.toLowerCase()] = target[addr]
+  })
```
sohkai: Super small nitpick: we could use `Object.entries` to directly get the target value for each key

packages/aragon-wrapper/src/utils/index.test.js
```diff
@@ -23,3 +23,18 @@ test('should enhance an object to lookup eth addresses easier', async (t) => {
   t.is(permissions['0x0000000000000000000000000000000000000abC'], undefined)
   t.is(permissions['0x0000000000000000000000000000000000000ABc'], undefined)
 })
+
+test('should initialise the proxy with its keys in lowercase', async (t) => {
```
sohkai: Perhaps we could rename this test to

`should allow the proxy to be initialized with an object containing any cased keys`

And add a few different variations (mixed case, all lower case).

packages/aragon-wrapper/src/cache/index.js
2color: Yeah I kind of felt that too

packages/aragon-wrapper/src/index.js
2color: Absolutely. In this case the move it quite minimal as it will go up two lines.

Issue #81, opened 2018-03-14
Implement caching on the filesystem


onbjerg:

sohkai:
From https://github.com/aragon/aragon.js/pull/238

> In general, we need a smart way to detect when the underlying chain changes (`--reset`) and invalidate the entire cache.

We can likely do this by detecting the block number (once we have block caching) and reseting it if it ever goes below what we've cached.

sohkai:
Closing this; caching on the filesystem is not very useful in light of other server-side solutions for reducing application state.

If one _really_ wanted to, they could write a wrapper above `@aragon/wrapper` that just persists the cache to a json file / etc.

PR #238, opened 2019-02-11
wrapper: use in-memory localforage driver on non-browser environments

sohkai:
Falls back to a non-persistent, in-memory cache in non-browser environments (e.g. node).

While this doesn't bring feature parity on non-browser environments, it at least avoids the "no storage" error and we don't yet have a use case for supporting caching in those other environments.

### DIFF
--- a/packages/aragon-wrapper/src/cache/index.js
+++ b/packages/aragon-wrapper/src/cache/index.js
@@ -22,6 +22,8 @@ export default class Cache {
     this.changes = new Subject()
 
     try {
+      // Make sure localforage has settled down and is not waiting for anything else
+      // before possibly setting new drivers
       await this.db.ready()
     } catch (err) {
       // If localforage isn't able to automatically connect to a driver
--- a/packages/aragon-wrapper/src/cache/index.js
+++ b/packages/aragon-wrapper/src/cache/index.js
@@ -9,34 +9,19 @@ import { concat } from 'rxjs/observable/concat'
 export default class Cache {
   constructor (prefix) {
     this.prefix = prefix
-
-    // Set up cache DB
-    this.db = localforage.createInstance({
-      driver: [localforage.INDEXEDDB, localforage.LOCALSTORAGE],
-      name: `localforage/${this.prefix}`
-    })
   }
 
   async init () {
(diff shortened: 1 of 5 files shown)


packages/aragon-wrapper/src/cache/index.js
```diff
@@ -9,22 +10,33 @@ export default class Cache {
...
       name: `localforage/${this.prefix}`
     })
   }
 
+  async init () {
+    // Set up the changes observable
+    this.changes = new Subject()
+
+    try {
+      await this.db.ready()
```
sohkai: This was really fun to debug; localforage has a race condition when there's no storage providers available if you try to set new drivers before waiting for `localforage.ready()`'s promise to resolve.

0x6431346e:
Added a commit to set the driver before initializing the db and to use the in-memory db for unit tests (wherever possible). What do you think?

sohkai:
See https://github.com/aragon/aragon.js/issues/81 on file caching.

Should be possible with a custom localforage driver, but worst case we can also just swap out localforage completely when on node environments.

packages/aragon-wrapper/src/cache/index.js
```diff
@@ -9,34 +9,19 @@ import { concat } from 'rxjs/observable/concat'
...
-      // use an in-memory driver instead
-      // TODO: this doesn't provide an persistent cache for node
-      if (this.db.driver() === null) {
-        await this.db.defineDriver(memoryStorageDriver)
-        await this.db.setDriver(memoryStorageDriver._driver)
-      }
-
-      await this.db.ready()
-    }
+    await localforage.defineDriver(memoryStorageDriver)
```
sohkai: @0x6431346e Oops, didn't see this yesterday!

What do you think about 8263657?

If we create the localforage instance first, we don't have to redefine the memory storage driver on subsequent instantiations (since the driver definitions are global to all localforage instances). I think it's also better to avoid defining the in-memory driver unless we absolutely have to.

0x6431346e: Oh, I see. Not as pretty, but makes sense if we optimize for performance :smile:

PR #240, opened 2019-02-21
wrapper: add ability to clear and remove items from cache

sohkai:
Useful when we need to reset app state, or the entire cache state (e.g. aragon/aragon's "Reset state" button).

**TODO**:

- [x] Tests

### DIFF
--- a/packages/aragon-wrapper/src/cache/index.js
+++ b/packages/aragon-wrapper/src/cache/index.js
@@ -3,12 +3,12 @@ import memoryStorageDriver from 'localforage-memoryStorageDriver'
 import { Subject } from 'rxjs/Rx'
 import { concat } from 'rxjs/observable/concat'
 
-const trackedKeys = new Set()
-
 /**
  * A cache.
  */
 export default class Cache {
+  #trackedKeys = new Set()
+
   constructor (prefix) {
     this.prefix = prefix
   }
@@ -64,7 +64,7 @@ export default class Cache {
   async clear () {
     await this.db.clear()
 
-    for (const key of trackedKeys) {
+    for (const key of this.#trackedKeys) {
       this.changes.next({ key, value: null })
     }
   }
@@ -78,7 +78,7 @@ export default class Cache {
    * @return {Observable}
    */
   observe (key, defaultValue) {
-    trackedKeys.add(key)
+    this.#trackedKeys.add(key)
 
     const keyChanges = this.changes
       .filter(change => change.key === key)
--- a/packages/aragon-wrapper/src/cache/index.js
+++ b/packages/aragon-wrapper/src/cache/index.js
@@ -3,6 +3,8 @@ import memoryStorageDriver from 'localforage-memoryStorageDriver'
 import { Subject } from 'rxjs/Rx'
 import { concat } from 'rxjs/observable/concat'
 
+const trackedKeys = new Set()
+
 /**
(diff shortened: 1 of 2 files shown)


packages/aragon-wrapper/src/cache/index.js
```diff
@@ -54,6 +54,19 @@ export default class Cache {
...
+    await this.db.removeItem(key)
+    this.changes.next({ key, value: null })
+  }
+
+  async clear () {
+    await this.db.clear()
+
+    // Reset the changes observable, since we've cleared the entire cache
+    this.changes.complete()
+    this.changes = new Subject()
```
sohkai: Curious what you guys think... this seems like the most obvious way to tell a user we're basically restarting the cache from scratch, but would also mean users would have to know about this and be re-subscribed.

Nothing is actually attached to these changes for now so it's not a big problem (perhaps this is observable is something we should just altogether remove?)

2color: I understand your concern. Seems reasonable to close the observable. 

+1 for removing this observable. Can we think of a reasonable use case for observing cache changing over time? 

This follows from the logic of using cache to improve performance ad hoc per request rather than per data change.

sohkai: Actually I lied; it's being used in `@aragon/client`: https://github.com/aragon/aragon.js/blob/master/packages/aragon-client/src/index.js#L180, and this is used in all the apps' frontends to subscribe to state changes in the apps.

Given that this is actually only observing a specific key, I think it actually makes more sense if we just send a `null` value back and not reset the observable. Moreover, `@aragon/client` subscriptions to this event would never know about the observable finishing in `@aragon/wrapper`'s context, since we don't transmit an end event over the `postMessage`.

2color: Fair enough. Yeah in that case we probably don't want to break this api.

Given the layers of abstraction and isolation in place, what do you think about cache being hidden away behind a simpler state call/subscription?

sohkai: > What do you think about cache being hidden away behind a simpler state call/subscription?

On some level this is already done for the apps (https://github.com/aragon/aragon.js/blob/master/packages/aragon-client/src/index.js#L180 returns an observable that emits on cache changes).

`aragon/aragon` doesn't use the cache very often, and `@aragon/wrapper` exposes it directly, so there's not much need for another abstraction.

packages/aragon-wrapper/src/cache/index.js
```diff
@@ -3,6 +3,8 @@ import memoryStorageDriver from 'localforage-memoryStorageDriver'
 import { Subject } from 'rxjs/Rx'
 import { concat } from 'rxjs/observable/concat'
 
+const trackedKeys = new Set()
```
2color: This will be shared across cache instances, which could leak change values when two instances are cleared. 

Any reason for not making this a class property?

sohkai: Ohh good point; mostly so I could hide it but let's make it a class property 👍

sohkai: Made it in a private class field in 940efbc9ea08b476f6e0ade49b79af38dca90b03 (just learned today that `babel@7.1.0` implemented this 🎉 )

PR #257, opened 2019-03-07
wrapper: deduplicate async requests when fetching app information

sohkai:
Adds a `AsyncRequestCache` that deduplicates async requests and caches the first successful result. This is really useful for getting information that will stay constant, such as contract values that can never change or contents from ipfs hashes.

Refactors the wrapper's `initApps()` to use this cache and also set up the stage for getting aragonPM repository information.

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -1,7 +1,7 @@
 // Externals
 import { ReplaySubject, Subject, BehaviorSubject, combineLatest, merge } from 'rxjs'
 import {
-  map, startWith, scan, tap, publishReplay, switchMap, flatMap, filter, first,
+  map, startWith, scan, tap, publishReplay, switchMap, filter, first,
   debounceTime, skipWhile
 } from 'rxjs/operators'
 import uuidv4 from 'uuid/v4'
@@ -28,6 +28,7 @@ import {
   makeProxy,
   makeProxyFromABI,
   getRecommendedGasLimit,
+  AsyncRequestCache,
   ANY_ENTITY
 } from './utils'
 
@@ -41,12 +42,6 @@ import Cache from './cache'
 
 // Interfaces
 import { getAbi } from './interfaces'
-
-// Cache for proxy values
-const proxyValuesCache = makeAddressMapProxy({})
-
-// Cache for app info, usually fetched from apm
-const appInfoCache = {}
 
 // Try to get an injected web3 provider, return a public one otherwise.
 export const detectProvider = () =>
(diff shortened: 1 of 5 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -348,59 +287,99 @@ export default class Aragon {
...
+                appInfo = await applicationInfoCache.request(cacheKey)
+              } catch (_) { }
             }
 
             return Object.assign(app, appInfo)
           })
         )
-      ),
-      // Replaying the last emitted value is necessary for this.apps' combineLatest to not rerun
-      // this entire operator chain on identifier emits (doing so causes unnecessary apm fetches)
```
sohkai: I think I was lying to myself when I made this comment before; `combineLatest()` only reduces emitted values so it should never rerun the entire operator chain. Maybe previously one of these observables wasn't hot (causing a new subscription chain to happen in the `combineLatest()`, but now all of them should be).

packages/aragon-wrapper/src/utils/AsyncRequestCache.js
```diff
@@ -0,0 +1,46 @@
...
+  request (key) {
+    if (this.has(key)) {
+      return this.#cache.get(key)
+    }
+    const request = Promise.resolve(this.#requestFn(key))
+      .catch((err) => {
+        this.#cache.delete(key)
+        throw err
+      })
+    this.#cache.set(key, request)
```
2color: This will cache the promise rather than the resolved value. Is that intentional?

2color: >  * A cache to deduplicate async requests.

My bad, this is to deduplicate requests of a key.

packages/aragon-wrapper/src/utils/AsyncRequestCache.test.js
```diff
@@ -0,0 +1,128 @@
...
+const wait = time =>
+  new Promise(resolve => {
+    setTimeout(resolve, time)
+  })
+
+test('AsyncRequestCache should cache requests', async (t) => {
+  // arrange
+  const requestKey = 'key'
+  const requestFn = sinon.spy(async key => {
+    wait(100)
```
2color: add `await` here?

packages/aragon-wrapper/src/utils/AsyncRequestCache.js
sohkai: Yes, the idea is to have a familiar interface, so that it _feels_ exactly the same as if you'd be calling it without the cache.

packages/aragon-wrapper/src/index.js
```diff
@@ -348,59 +287,99 @@ export default class Aragon {
...
+        const appsWithoutKernel = proxyAddresses.filter((address) => !this.isKernelAddress(address))
         return [this.kernelProxy.address].concat(appsWithoutKernel)
       }),
       // Get proxy values
       switchMap(
-        (appAddresses) => Promise.all(
-          appAddresses.map((app) => this.getProxyValues(app))
+        (proxyAddresses) => Promise.all(
+          proxyAddresses.map(async (proxyAddress) => {
+            let proxyValues
```
2color: Can we simplify this step by adding the `proxyAddress` to the object constructed in the `requestFn` of `proxyContractValueCache`?

packages/aragon-wrapper/src/index.js
```diff
@@ -348,59 +287,99 @@ export default class Aragon {
...
         )
-      ),
-      // Replaying the last emitted value is necessary for this.apps' combineLatest to not rerun
-      // this entire operator chain on identifier emits (doing so causes unnecessary apm fetches)
-      publishReplay(1)
+      )
     )
-    this.appsWithoutIdentifiers.connect()
 
+    this.identifiers = new Subject()
```
2color: What are `identifiers`?

packages/aragon-wrapper/src/utils/AsyncRequestCache.js
2color: Yeah makes total sense. It's also interoperable with just values (as in the case of `applicationInfoCache` calling `getAragonOsInternalAppInfo`) which is nice

packages/aragon-wrapper/src/index.js
sohkai: I like the current implementation a bit better, as it'll ensure `proxyAddress` is added even if the request fails. We'd otherwise always have to resolve in `proxyContractValueCache`, which would break the cache invalidation mechanism on fails.

packages/aragon-wrapper/src/index.js
sohkai: These are the "app identifiers" (called by [`app.identify` in the API](https://github.com/aragon/aragon.js/blob/master/packages/aragon-api/src/index.js#L68)), that allow an app to "identify" itself (e.g. with a particular token instance, or voting params).

PR #259, opened 2019-03-12
Add @aragon/api-react

bpierre:
# aragonAPI for React

This module allows to interact with aragonAPI using [React Hooks](https://reactjs.org/docs/hooks-intro.html). [`@aragon/api`](https://github.com/aragon/aragon.js/blob/master/docs/APP.md) is used under the hood, so being familiar with it can be useful.

## Usage

```jsx
import { AragonApi, useAragonApi } from '@aragon/api-react'

function App() {
  const { api, appState } = useAragonApi()
  const { count = 0 } = appState

  return (
    <div>
      <div>{count}</div>
      <button onClick={() => api.increment(1)}>
        Increment
      </button>
    </div>
  )
}

ReactDOM.render(
  <AragonApi>
    <App />
  </AragonApi>,
  document.getElementById('root')
)
```

This is a simple example demonstrating how we can use aragonAPI for React to connect the app to its contract, fetch some data from its state (using `appState`), and trigger an action on it (with `api.increment(1)`). The full API is detailed below.

### &lt;AragonApi />

Before using any Hook provided, you need to declare this component to connect the app. It is generally a good idea to do it near the top level of your React tree. It should only be declared once.

It has an optional `reducer` prop, which lets you process the state coming from the [background script](https://hack.aragon.org/docs/aragonjs-guide-bg-scripts.html). If not provided, the state is passed as is.

#### Example

```jsx
import { AragonApi, useAppState } from  '@aragon/api-react'
import BN from 'bn.js'

function App() {
  const { balance } = useAppState()
  return (
    <div>{balance.toString(10)}</div>
  )
}

function reducer(state) {
  if (state === null) { // initial sync
    return { balance: new BN(0) }
  }
  return { balance: new BN(state.balance) }
}

ReactDOM.render(
  <AragonApi reducer={reducer}>
    <App />
  </AragonApi>,
  document.getElementById('root')
)
```

### useAragonApi()

A React Hook that returns the data needed to interact with the app contract.

As with any React Hook, please ensure that you follow the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html).

It returns an object containing the following entries:

#### `api`

This is the current [`AragonApp`](https://github.com/aragon/aragon.js/blob/master/docs/APP.md#aragonapp) instance. Use it to call methods on the contract.

Example:

```jsx
function App() {
  const { api } = useAragonApi()
  return (
    <button onClick={() => api.vote(true)}>Vote</button>
  )
}
```

#### `appState`

The app state, after having passed the [background script](https://hack.aragon.org/docs/aragonjs-guide-bg-scripts.html) state through the `reducer` prop of `AragonApi`.

Example:

```jsx
import { useAragonApi } from  '@aragon/api-react'

function App() {
  const { appState } = useAragonApi()
  return (
    <div>{appState.count}</div>
  )
}
```

#### `connectedAccount`

The connected Ethereum account. Its value is `""` (empty string) when there is no account connected.

Example:

```jsx
function App() {
  const { connectedAccount } = useAragonApi()
  return (
    <div>Account: {connectedAccount? connectedAccount : 'Not connected'}</div>
  )
}
```

#### `network`

An [object](https://github.com/aragon/aragon.js/blob/master/docs/APP.md#network) representing the current network using its `id` and `type` entries. Its value is `null` until it gets loaded.

Example:

```jsx
function App() {
  const { network } = useAragonApi()
  return (
    <div>Current network: {network.type}</div>
  )
}
```

#### `displayMenuButton`

Whether or not to display the menu button (`Boolean`), depending on it being automatically hidden or not in the client.

#### `requestMenu()`

Call this function to display the Aragon menu, when hidden automatically. This should be called when the user clicks on the menu button.

### useApi()

This Hook returns the same data than the `api` entry from the `useAragonApi()` hook.

### useAppState()

This Hook returns the same data than the `appState` entry from the `useAppState()` hook.

### useConnectedAccount()

This Hook returns the same data than the `connectedAccount` entry from the `useAragonApi()` hook.

### useMenuButton()

This Hook returns an array containing the `displayMenuButton` and the `requestMenu` entries from the `useAragonApi()` hook, in that order.

### useNetwork()

This Hook returns the same data than the `network` entry from the `useAragonApi()` hook.

## License

AGPL-3.0


bpierre:
@0x6431346e Thanks!

> I'm not sure I understand the use case of the second reducer though (the one passed to useAragonApi)🤔.

This is the reducer that transforms the data coming from the RxJS observer into a data structure ready to be used by the React app. By keeping this part outside the rendering function, it allows to process emitted values only once. It is entirely optional. You can see an example of this kind of reducer in the [Voting app](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/App.js#L274-L334) or the [Token Manager app](https://github.com/aragon/aragon-apps/blob/master/apps/token-manager/app/src/App.js#L170-L215).

> Is this hook expected to be used only once, in the 'App'/top-level component and the api to be passed down to children or would each child use the hook?

For the initial version, we decided that it would be better to make it a hook for the top level component only, to avoid any confusion around the passed reducer… but I agree this might be a bit confusing.

A solution could be to have `useAragonApi()` work as a read-only hook when it gets used by a child:

```jsx
import { useAragonApi } from '@aragon/api-react'

const appStateReducer = state => ({ ...state, /* … */ })

// The initial call would initiate the connection,
// and be the only call allowed to update the reducer.
const App = () => {
  useAragonApi(appStateReducer)
  return <Child />
}

// This would work fine…
const Child = () => {
  const { appState } = useAragonApi()
  return <p>{appState.value}</p>
}

// …but this would trigger an error.
const Child = () => {
  const { appState } = useAragonApi(appStateReducer)
  return <p>{appState.value}</p>
}
```

It should be feasible using `useEffect()` (to ensure only one component is using the initial version), but I don’t like having a function with variable parameters depending on its context, it might feel even more confusing.

Another solution could be to provide another hook, or a set of hooks, to retrieve this data:

```jsx
import { useAragonApi, useAragonApiContext } from '@aragon/api-react'

const appStateReducer = state => ({ ...state, /* … */ })

// This doesn’t change.
const App = () => {
  useAragonApi(appStateReducer)
  return <Child />
}

// But the children would have to use another hook:
const Child = () => {
  const { api, appState } = useAragonApiContext()
  return <p>{appState.value}</p>
}
```

```jsx
// We could also export several hooks:

import {
  useApi,
  useAppState,
  useNetwork,
  useConnectedAccount,
  useMenu,
} from '@aragon/api-react'

/* … */

const Child = () => {
  const api = useApi()
  const appState = useAppState()
  const network = useNetwork()
  const { displayMenuButton, requestMenu } = useMenu()
  return <p>{appState.value}</p>
}
```

We could also decide to separate the initialization from the data retrieving, perhaps by renaming `useAragonApi` into something like `useConnectApp` (or anything more elegant than having two verbs next to each other 😆):


```jsx
import {
  useApi,
  useAppState,
  useConnectApp,
} from '@aragon/api-react'

const appStateReducer = state => ({ ...state, /* … */ })

const App = () => {

  // required to make the others hooks return something
  useConnectApp(appStateReducer)

  // these can be called from anywhere in the tree
  const api = useApi()
  const appState = useAppState()

  return <p>{appState.value}</p>
}
```

@sohkai @0x6431346e thoughts?

sohkai:
> For the initial version, we decided that it would be better to make it a hook for the top level component only, to avoid any confusion around the passed reducer… but I agree this might be a bit confusing.

My original thinking was to force users to only use this once, and they could handle any sharing of the state internally. However, I'm starting to think this could be confusing since hooks seem like they should be reusable anywhere in the render tree...

We could make it possible for any call to `useAragonApi()` to connect a reducer, but this would break most of the current hook. I don't want to be creating multiple `Aragon` objects every time a component uses the hook, for example; it'd be much nicer to share the both the `Aragon` object, as well as its observables' state.

It then sounds like either `useAragonApiContext()` or `useConnectApp()` would work, to create a provider-consumer pairing. But this would also mean more boilerplate code, and asks the question of where we would use the provider hook.

2color:
> It then sounds like either useAragonApiContext() or useConnectApp() would work, to create a provider-consumer pairing. But this would also mean more boilerplate code, and asks the question of where we would use the provider hook.

Not sure I understood.

0x6431346e:
> This is the reducer that transforms the data coming from the RxJS observer into a data structure ready to be used by the React app. By keeping this part outside the rendering function, it allows to process emitted values only once. It is entirely optional. You can see an example of this kind of reducer in the Voting app or the Token Manager app.

Ah, I see, makes sense.

> For the initial version, we decided that it would be better to make it a hook for the top level component only, to avoid any confusion around the passed reducer… but I agree this might be a bit confusing.

> My original thinking was to force users to only use this once, and they could handle any sharing of the state internally.

Would be awesome to give a warning/error if there's an easy way to detect this.

> We could also decide to separate the initialization from the data retrieving, perhaps by renaming useAragonApi into something like useConnectApp (or anything more elegant than having two verbs next to each other 😆):

Sounds like a great v2.

```
// This would work fine…
const Child = () => {
  const { appState } = useAragonApi()
  return <p>{appState.value}</p>
}
```
In this case the Child would render on any state change, right?
Maybe we could do something like this: https://github.com/facebookincubator/redux-react-hook#usemappedstatemapstate

sohkai:
@2color 

> It then sounds like either useAragonApiContext() or useConnectApp() would work, to create a provider-consumer pairing. But this would also mean more boilerplate code, and asks the question of where we would use the provider hook.

> Not sure I understood.

If we are to share the top level state (the `Aragon` object as well as a number of observables), we'll need a way to provide that into the other hooks (used by children), hence the provider-consumer pairing.

If we split this up, then we'll have to use the provider hook separately, and figure out which component it belongs in. In the example given for `useConnectApp()` though, it seems like we could actually do this in the same component, so maybe it's not a problem.

2color:
> If we are to share the top level state (the Aragon object as well as a number of observables), we'll need a way to provide that into the other hooks (used by children), hence the provider-consumer pairing. useConnectApi() would be the provider, and the other hooks would be consumers.

I undertand. 

>  I don't want to be creating a new Aragon object every time a component uses the hook

👍

> We could also decide to separate the initialization from the data retrieving, perhaps by renaming useAragonApi into something like useConnectApp 

Separating/splitting seems logical. We certainly want to ensure initialisation happens only once.

> hence the provider-consumer pairing. useConnectApi() would be the provider, and the other hooks would be consumers

Implementation question, would this rely on react context that'd be used in our hook implementation. Basically just abstracting the details of the context provider/consumer pattern?

bpierre:
Thank you all for the feedback, it’s been super useful!

@0x6431346e 

> ```jsx
> // This would work fine…
> const Child = () => {
>   const { appState } = useAragonApi()
>   return <p>{appState.value}</p>
> }
> ```
>
> In this case the Child would render on any state change, right?

Yes 👍 

> Maybe we could do something like this: https://github.com/facebookincubator/redux-react-hook#usemappedstatemapstate

Thanks, this module inspired me to propose a new solution for the API, that I like much better! 👌

Here are the changes:

- There is now a `<ConnectAragonApp />` component that need to be declared at the top level of the app. This component is responsible of connecting the app and takes an optional `reducer` prop. It can only be declared once in an app. If any hook is used without having this component declared at an upper level, an error is thrown. The `reducer` prop can be updated if needed.

- The `useAragonApi()` hook is now only a consumer: it can be used anywhere, as many time as we want, and doesn’t take any parameter.

- The render props API has been removed for clarity, and because it is now very simple to implement if needed: `({ children }) => children(useAragonApi())`.
- Add alias hooks for convenience: `useNetwork()`, `useApi()`, etc. (see README.md)

@2color @0x6431346e @sohkai what do you think?

packages/aragon-api-react/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+function App() {
+  const { appState } = useAragonApi(reducer)
+  return (
+    <div>{appState.balance.toString(10)}</div>
+  )
+}
+
+ReactDOM.render(
+  <ConnectAragonApi
```
AquiGorka: Would it make sense to offer a [_redux-HOC-like-connect_](https://react-redux.js.org/api/connect) too?

packages/aragon-api-react/README.md
```diff
@@ -0,0 +1,173 @@
...
+It has an optional `reducer` prop, which lets you process the state coming from the background script. If not provided, the state is passed as is from the background script.
+
+### Example
+
+```jsx
+import BN from 'bn.js'
+import { ConnectAragonApi, useAragonApi } from  '@aragon/api-react'
+
+function App() {
+  const { appState } = useAragonApi(reducer)
```
2color:
```suggestion
  const { appState } = useAragonApi()
```

packages/aragon-api-react/README.md
```diff
@@ -0,0 +1,173 @@
...
+  >
+    <App />
+  </ConnectAragonApi>,
+  document.getElementById('root')
+)
+```
+
+
+
+## `useAragonApi(reducer)`
```
2color:
```suggestion
## `useAragonApi()`
```

packages/aragon-api-react/README.md
```diff
@@ -0,0 +1,173 @@
...
+function App() {
+  const { appState } = useAragonApi(reducer)
+  return (
+    <div>{appState.balance.toString(10)}</div>
+  )
+}
+
+ReactDOM.render(
+  <ConnectAragonApi
+    reducer={state => {
```
2color: What do you think about declaring the reducer function outside the component?

This is based on the documentation patterns I see in https://redux.js.org/advanced/example-reddit-api

2color:
Nice! The latest changes make this really clear and elegant. 

Regarding @AquiGorka's suggestion (`export default ConnectAragonApi(App, reducer)`), I'm aware that HOCs (Higher order components) are [contending with render props](https://gist.github.com/heygrady/f9bf3b6dd93fe3d87ba87430fd3c20d5) as a solution for some problems.

@bpierre What do you think about this?

packages/aragon-api-react/README.md
bpierre: Yes let’s do this 👍 I was trying to make it as compact as possible for the documentation, but it looks better outside of it.

bpierre:
@AquiGorka @2color Regarding having a function: would it be an HOC in this case, since we don’t pass any data to the wrapped component? I see it more as an [“enhanced” context provider](https://github.com/aragon/aragon.js/pull/259/files#diff-df3ed8cb308cbe434aac1d4a00e05543R94).

But I understand the term `connectSomething()` (and `withSomething()`) is often used in the React world for HOCs (as in, “a function that returns a new component with implicit props passed to it”), so the term might be a bit confusing, even though it refers to “initiate the connection to aragonAPI” here, and not “connect this component to something”. Maybe we could rename it `<AragonApiProvider>`?

It would be similar to the [`Provider`](https://react-redux.js.org/api/provider) component of React Redux, which behaves like the `AragonApiConnect` component.

I think a good case for HOC functions would be for the `useSomething` hooks we provide, but hooks were invented to solve this exact problem 😆

What do you guys think?

2color:
On a second thought, I think it's sensible as you have it right now.

bpierre:
I start to like `AragonApiProvider` as it might be a bit clearer… and I was wondering if `AragonApi` could work to, now that the render prop that was using this name got removed? Or if it’s too weird to also have `useAragonApi()` doing something different. 🤔

We could also name it `Provider` like React Redux does, but for named exports I generally prefer less generic names.

bpierre:
> Regarding the hoc thing, I agree it is more an "enhancer", my original question was if it would make sense to offer such enhancer as one other way to "connect" mostly because of the similarity to redux's connect.

I would prefer not to, because its behavior is closer to the `Provider` of React Redux than its `connect()` function, even though the reducer is passed to the provider in our case.

Making it possible to declare the provider using a function named `connect()` could let people think that it works like React Redux, and can be used to connect any component to the store, while it is not the case: it should only be declared once, at the top level.

What is similar to the React Redux `connect()` is the collection of hooks itself, exactly the same way `redux-react-hook` removed the `connect()` function to provide [`useMappedState(reducer)`](https://github.com/facebookincubator/redux-react-hook#usemappedstatemapstate) instead.

We could also add a function like `provideAragonApi()`, but I’m not sure why that would be needed (compared to declaring a component)? I think we should aim to provide “[one obvious way to do something](https://www.python.org/dev/peps/pep-0020/)”, when possible. 😆

packages/aragon-api-react/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+This is a simple example demonstrating how we can use aragonAPI for React to connect the app to its contract, fetch some data from its state (using `appState`), and trigger an action on it (with `api.increment(1)`). The full API is detailed below.
+
+## Documentation
+
+### &lt;AragonApi />
+
+Before using any Hook provided, you need to declare this component to connect the app. It is generally a good idea to do it near the top level of your React tree. It should only be declared once.
+
+It has an optional `reducer` prop, which lets you process the state coming from the background script. If not provided, the state is passed as is from the background script.
```
sohkai: Might be nice to link to https://hack.aragon.org/docs/aragonjs-guide-bg-scripts.html or https://github.com/aragon/aragon.js/blob/master/docs/BACKGROUND_SCRIPTS.md

packages/aragon-api-react/yarn.lock
```diff
@@ -0,0 +1,2128 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
sohkai: Let's remove this for now, until we have a coherent lock file strategy :).

packages/aragon-api-react/src/index.js
```diff
@@ -0,0 +1,150 @@
...
+const AragonApiContext = createContext()
+
+function AragonApi({
+  children,
+  reducer = state => (state === null ? {} : state),
+  onMessage = () => {},
+}) {
+  const [api, setApi] = useState(null)
+  const [connectedAccount, setConnectedAccount] = useState('')
+  const [network, setNetwork] = useState('')
```
sohkai: Should this also start off as `null` or an empty object?

packages/aragon-api-react/src/index.js
```diff
@@ -0,0 +1,150 @@
...
+            .pipe(map(state => reducer(state)))
+            .subscribe(state => setAppState(state)),
+
+          // account
+          api
+            .accounts()
+            .subscribe(accounts => setConnectedAccount(accounts[0] || '')),
+
+          // network
+          api.network().subscribe(network => setNetwork(network || '')),
```
sohkai: Don't think the default `''` makes sense, since this should be an object

packages/aragon-api-react/src/index.js
```diff
@@ -0,0 +1,150 @@
...
+
+    window.addEventListener('message', handleMessage)
+
+    return () => {
+      window.removeEventListener('message', handleMessage)
+      subscribers.forEach(subscriber => subscriber.unsubscribe())
+    }
+  }, [api, reducer])
+
+  return createElement(
```
sohkai: Was there a particular reason to use `createElement()`? Just syntax?

packages/aragon-api-react/src/index.js
bpierre: Yes, good catch!

packages/aragon-api-react/src/index.js
bpierre: It was to avoid adding `babel-preset-react` only for one `createElement` call. Do you think we should use the JSX syntax?

sohkai: 👍 Nah, but let's leave a comment for this :).

PR #266, opened 2019-03-22
Wrapper: don't assign initializationBlock on non-kernel proxies when unneeded

sohkai:
@2color Wondering what you think might be the _least_ confusing.

Currently, when we make these contract proxies, we use the Kernel's `initializationBlock` as their `initializationBlock` (this has large implications when fetching events, because we default to it a the `fromBlock`).

However, this is _technically_ the incorrect block for app proxies, as each of these proxies can have their own `initializationBlock`, and we don't actually need it to be set in changeset as we never fetch events there.

I don't envision us ever having to fetch events here, so I thought it might be clearer to just not set the `initializationBlock`.

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -296,7 +296,7 @@ export default class Aragon {
 
     const proxyContractValueCache = new AsyncRequestCache((proxyAddress) => {
       if (this.isKernelAddress(proxyAddress)) {
-        const kernelProxy = makeProxy(proxyAddress, 'ERCProxy', this.web3, this.kernelProxy.initializationBlock)
+        const kernelProxy = makeProxy(proxyAddress, 'ERCProxy', this.web3)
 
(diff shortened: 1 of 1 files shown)


2color:
Not directly related to this PR but can you shed light on why there are two similar – albeit with a different interface – instances of the `kernelProxy`:
https://github.com/aragon/aragon.js/blob/4914b592f85de80388959e2f103749f1e5da7f11/packages/aragon-wrapper/src/index.js#L137-L139

and https://github.com/aragon/aragon.js/blob/4914b592f85de80388959e2f103749f1e5da7f11/packages/aragon-wrapper/src/index.js#L303

sohkai:
> Not directly related to this PR but can you shed light on why there are two similar – albeit with a different interface – instances of the kernelProxy:

Yes, the ABIs are different for those two. `Kernel` is the base interface, whereas `ERCProxy` is the proxy interface.

PR #267, opened 2019-03-22
Wrapper: Handle SetApp for updated apps

sohkai:
[`kernel.SetApp()`](https://github.com/aragon/aragonOS/blob/dev/contracts/kernel/Kernel.sol#L217) is emitted when a new app is installed or updated (for all instances of the app).

We don't need to detect old emissions of this event, as we already fetch the latest values of the app proxies on initialization. However, we do need to listen to new events, as an app could be updated at run time.

This adds a subscription that will emit a new list of apps on `this.apps`, with any updated apps having an `updated: true` flag that will allow clients to detect and handle (e.g. stop and rerun any WebWorkers, etc.).

**This is a breaking change.**

--------

_Note_: I've chosen "updated" as the terminology instead of "upgraded" as technically you could also be downgrading an app when this event is emitted.

- [x] Test against local chain
- [x] Update client to detect and swap out updated apps
- [x] Should probably write a test that includes mocked event emissions on `SetApp`


packages/aragon-wrapper/src/core/proxy/index.test.js
```diff
@@ -51,13 +51,119 @@ test('should get only the requested events', (t) => {
...
+  events.subscribe(event => {
+    if (eventCount === 0) {
+      t.deepEqual(event.amount, 5)
+    } else if (eventCount === 1) {
+      t.deepEqual(event.amount, 10)
+    } else {
+      // Should only see it twice
+      t.fail()
+    }
+    ++eventCount
```
sohkai: Not sure if there's a better way to write this type of test when multiple emissions are expected.

packages/aragon-wrapper/src/index.js
```diff
@@ -438,6 +464,36 @@ export default class Aragon {
...
+   * @param {string} identifier The identifier of the app
+   * @return {void}
+   */
+  setAppIdentifier (address, identifier) {
+    this.identifiers.next({
+      address,
+      identifier
+    })
+  }
+
```
sohkai: Note that these changes to the `identifiers` means that this is a breaking change, as the identifier and app lists are not merged by default anymore.

I hadn't realized that the previous logic could create infinite loops until I actually encountered them:

1. Init apps, get apps observable
1. Client starts apps
1. Apps emit on the identifier observable
1. Due to the `combineLatest()` logic, another apps emission would occur
1. If the client then restarts the app due to the apps emission (e.g. in the case of an upgrade), the app could re-emit its identifier
1. Repeat from 4

Hence, I've taken the easy way out by splitting the two concepts from each other and pushing the work to clients.

sohkai: Since it's a breaking change, I also took the liberty of renaming `identifier` to `appIdentifier` to make it more clear at a glance what this observable is for.

packages/aragon-wrapper/src/index.js
```diff
@@ -331,6 +340,15 @@ export default class Aragon {
     // Get all app proxy addresses
     const baseApps$ = this.permissions.pipe(
       map(Object.keys),
+      // Dedupe until apps change
+      distinctUntilChanged((oldProxies, newProxies) => {
+        if (oldProxies.length !== newProxies.length) {
+          return true
```
2color: Shouldn't this be false? The comparison function returns a bool of whether two items are the same (based on the example -https://rxjs-dev.firebaseapp.com/api/operators/distinctUntilChanged)

2color: It's badly documented, but based on the example the comparison functions is `isEqual` rather than `isDistinct`

packages/aragon-wrapper/src/index.js
```diff
@@ -376,8 +377,51 @@ export default class Aragon {
       ))
     )
 
+    // SetApp events are emitted when apps are installed and upgraded
+    // These may modify the implementation addresses of the proxies (modifying their behaviour), so
+    // we invalidate any caching we've done
+    const updatedApps$ = this.kernelProxy
+      // Override events subscription with empty options to subscribe from latest block
+      .events('SetApp', {})
```
2color: If `fromBlock` is not sent on calls to `events`, does it only emit events from the latest block? 

It's not documented in https://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#id34

packages/aragon-wrapper/src/index.js
```diff
@@ -376,8 +377,51 @@ export default class Aragon {
...
+              return {
+                ...app,
+                ...proxyValues,
+                updated: true
+              }
+            })
+          }
+        ),
+        // Switch to resolved array of promises
+        switchMap(Promise.all)
```
2color: The `switchMap` here will call Promise.all on every new array of promises returned from the previous operator (`withLatestFrom`) and emit the resolved values of `Promise.all`. Why however do we use `switchMap` over `mergeMap`?

Can't we have a situation in which two `SetApp` events are emitted and we want the changes of both?

packages/aragon-wrapper/src/index.js
sohkai: Yes, this is more or less buried in the code. It assumes latest if you don't include it or set it to `0`.

See [latest from master](https://github.com/ethereum/web3.js/blob/cd32de24bf/packages/web3-core-subscriptions/src/subscriptions/eth/LogSubscription.js#L53), and [the version we use (v1.0.0-beta.34)](https://github.com/ethereum/web3.js/blob/fcf6997d15/packages/web3-core-subscriptions/src/subscription.js#L222).

Made an issue for it: https://github.com/ethereum/web3.js/issues/2668.

packages/aragon-wrapper/src/index.js
sohkai: > Can't we have a situation in which two SetApp events are emitted and we want the changes of both?

Hmm, yes, this is true. I had thought that `mergeMap()` and `switchMap()` are synonymous when it comes to single-emission observables being returned from their operators, but that's not true when the first event is still waiting to get processed when the second event fires.

There are still cases where you'd only be interested in the latest event, e.g. how the `appIdentifiers` are set up, but I think all our uses of `switchMap()` should be switched to `mergeMap()` 👍.

packages/aragon-wrapper/src/index.js
sohkai: Good catch! Fixed in e6d3afc.

packages/aragon-wrapper/src/index.js
sohkai: See eaabdec, the decision to use `switchMap()` or `mergeMap()`, as I understand it, is if we care about the throttled values.

packages/aragon-wrapper/src/index.test.js
```diff
@@ -360,69 +360,89 @@ appInitTestCases.forEach(([testName, permissionsObj]) => {
...
+test('should init the app identifiers correctly', async (t) => {
+  t.plan(1)
+  // arrange
+  const { Aragon } = t.context
+  const instance = new Aragon()
+  // act
+  await instance.initAppIdentifiers()
+  // assert
+  instance.identifiers.subscribe(value => {
+    t.deepEqual(value, {})
```
2color: What's the reason that we emit an empty object initially? Is that due to the consumers rely on there always being some value (at times an empty object like here)

sohkai: No particular reason, but an empty one seems reasonable to begin with, API-wise, since the app identifier mapping should alway "exist" (but may not have any identifiers declared inside it).

packages/aragon-wrapper/src/index.js
sohkai: So I figured this one out; the one we usually want with promises is actually [`concatMap()`](https://rxjs-dev.firebaseapp.com/api/operators/concatMap) because we'd like to queue each promise's resolution behind each other (rather than firing at will or throttling).

PR #268, opened 2019-03-22
Wrapper: add installedRepos observable

sohkai:
Adds an `installedRepos` observable that emits an array of `InstalledRepos`:

```js
// InstalledRepo:
{
  repoAddress: '0x...',
  appId: '0x...',
  currentVersion: {
    version: '1.1.1',
    content: {...},
  },
  latestVersion: {
    version: '2.0.0',
    content: {...},
  },
  versions: [
    { version: '1.0.0', contractAddress, contentURI, timestamp },
    {...}
  ]
}
```

Eschews using [`apm.js`](https://github.com/aragon/apm.js/blob/master/src/index.js)'s utilities for more optimized local versions of them due to a few design limitations in `apm.js`:

1. Returns raw `Web3.Contract`s, so we'd have to wrap its events around an Observable (we already have the `Proxy` to do this)
1. Most calls attempt to return the full app info from IPFS (`artifact.json` + `manifest.json`), so they can take quite long to resolve
1. A lot of unnecessary duplicated requests

-------

TODO:

- [x] Actually test against the client
- [x] Fix tests
- [ ] Write new tests for `installedRepos`

We don't currently document the exposed observables, but we should!

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -31,7 +31,7 @@ import Messenger from '@aragon/rpc-messenger'
 import * as handlers from './rpc/handlers'
 
 // Utilities
-import { makeRepoProxy, getAllVersionsOfRepo } from './core/apm'
+import { makeRepoProxy, getAllRepoVersions, getRepoVersionById } from './core/apm'
 import { getAragonOsInternalAppInfo, getKernelNamespace } from './core/aragonOS'
 import { CALLSCRIPT_ID, encodeCallScript } from './evmscript'
 import {
@@ -454,6 +454,12 @@ export default class Aragon {
       )
     )
 
+    this.apps = appsWithInfo$.pipe(
+      publishReplay(1)
+    )
+    this.apps.connect()
+
+    // Initialize installed repos from the list of apps
     this.installedRepos = new ReplaySubject(1).pipe(
       // Reduce single repo version emissions over time into one expanding array of installed repos
       scan((repos, updatedRepo) => {
@@ -470,12 +476,16 @@ export default class Aragon {
     )
 
     apps$.pipe(
-      // Map apps into a deduped list of apm repos, with these assumptions:
+      // Map installed apps into a deduped list of their apm repos, with these assumptions:
       //   - No apps are lying about their appId (malicious apps _could_ masquerade as other
       //     apps by setting this value themselves)
       //   - `contractAddress`s will stay the same across all installed apps.
       //      This is technically not true as apps could set this value themselves
       //      (e.g. as pinned apps do), but these apps wouldn't be able to upgrade anyway
+      //
+      //  Ultimately returns an array of objects, holding the repo's:
+      //    - appId
+      //    - base contractAddress
       map((apps) => Object.values(
         apps.reduce((apmRepos, app) => {
           const { appId, contractAddress, isAragonOsInternalApp } = app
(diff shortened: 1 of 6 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -439,6 +439,109 @@ export default class Aragon {
       )
     )
 
+    this.installedRepos = new ReplaySubject(1).pipe(
```
sohkai: `initApps()` is becoming way too much of a beast.

I'd like to start exploring how we can compartmentalize this into its own module; one idea:

- Perhaps this module could expose one big `init()` function (or multiple smaller `initX()` functions), that returns these observables we attach onto `this`
- This module could be built up of multiple smaller methods that don't rely on the global `this` state

packages/aragon-wrapper/src/index.js
```diff
@@ -450,6 +469,276 @@ export default class Aragon {
...
+     *                             *
+     ******************************/
+
+    // Initialize installed repos from the list of apps
+    const installedRepoCache = new Map()
+    const repo$ = apps$.pipe(
+      // Map installed apps into a deduped list of their aragonPM repos, with these assumptions:
+      //   - No apps are lying about their appId (malicious apps _could_ masquerade as other
+      //     apps by setting this value themselves)
+      //   - `contractAddress`s will stay the same across all installed apps.
```
2color: What is `contractAddress`? how could it stay the same for different apps (respectively different APM repo)?

packages/aragon-wrapper/src/index.js
```diff
@@ -450,6 +469,276 @@ export default class Aragon {
...
+      //      This is technically not true as apps could set this value themselves
+      //      (e.g. as pinned apps do), but these apps wouldn't be able to upgrade anyway
+      //
+      //  Ultimately returns an array of objects, holding the repo's:
+      //    - appId
+      //    - base contractAddress
+      map((apps) => Object.values(
+        apps
+          .filter(({ appId }) => !isAragonOsInternalApp(appId))
+          .reduce((installedRepos, { appId, codeAddress, updated }) => {
```
2color: Here we reduce the apps array of objects into an object which is then transformed into an array again by the wrapping `Object.values` call. Can't we just map the array after filter directly into the format without the transformation into an object?

sohkai: The reason we map back and forth between the array and the object is to dedupe the array based on the `appId`, as it may include multiple instances of an app (e.g. the AGP org has two voting app instances).

This could be combined with the following `map()` operator, which may be more clear 🤷‍♂️.

packages/aragon-wrapper/src/index.js
sohkai: Yeah, `codeAddress` and `contractAddress` are almost synonymous terms that could be explained a bit more.

- `codeAddress` is used by the **proxy** contract, to describe where its implementation lives
- `contractAddress` is used by the **Repo** contract, to declare a specific version's on-chain package location

A **proxy**'s `codeAddress` should always be pointing (linking) to a **Repo** version's `contractAddress`. An upgrade to the **proxy** will modify its `codeAddress` to point to another version's `contractAddress`.

packages/aragon-wrapper/src/index.js
```diff
@@ -457,6 +475,276 @@ export default class Aragon {
...
+                }
+              }),
+
+              // Trick to immediately emit (e.g. similar to a do/while loop)
+              startWith(null),
+
+              // Reduce newly emitted versions into the full list of versions
+              scan(({ appId, versions }, newVersionInfo) => {
+                let newVersions = versions
+                if (newVersionInfo) {
```
2color: is `newVersionInfo` ever falsey ?

sohkai: Directly preceding it, with the [`startsWith(null)`](https://github.com/aragon/aragon.js/pull/268/files/e02f0f1f5092fab2cf557dae1d6e920f0423d5c6#diff-5b914b5b526a57f89974fa2fe009ee80R620).

packages/aragon-wrapper/src/index.js
```diff
@@ -457,6 +475,276 @@ export default class Aragon {
...
+                  const versionIndex = versions.findIndex(({ versionId }) => versionId === newVersionInfo.versionId)
+
+                  if (versionIndex === -1) {
+                    newVersions = versions.concat(newVersionInfo)
+                  } else {
+                    newVersions = Array.from(versions)
```
2color: Is this the case where we process the `NewVersion` event for an existing version that came from `initialVersions`?

packages/aragon-wrapper/src/index.js
```diff
@@ -457,6 +475,276 @@ export default class Aragon {
...
+                    newVersions[versionIndex] = newVersionInfo
```
2color: Is this the case where we process the `NewVersion` event for an existing version that came from `initialVersions`?

packages/aragon-wrapper/src/index.js
2color: is the [`startsWith(null)`](https://github.com/aragon/aragon.js/pull/268/files/e02f0f1f5092fab2cf557dae1d6e920f0423d5c6#diff-5b914b5b526a57f89974fa2fe009ee80R620) to flush the pipeline with whatever we already have in `initialVersions`?  In

packages/aragon-wrapper/src/index.js
2color: In otherwords, is this to decorate/enhance with information from `newVersionInfo` (seems like `timestamp` is the only addition)

packages/aragon-wrapper/src/index.js
sohkai: Yup, exactly, so we don't need to wait for the `eth_getlogs` to finish

packages/aragon-wrapper/src/index.js
sohkai: Yes, it's basically to decorate old versions with the `timestamp`, and add new versions if they come while the app is open.

PR #270, opened 2019-03-22
@aragon/api-react: remove the rxjs dependency

bpierre:
Builds on  #269 − please don’t merge before it gets merged.

- Remove the `rxjs` dependency.
- Move `@aragon/api` as a peerDependency.
- Compress the build with Terser.
- Disable `webpack` in the size-limit config for `@aragon/api-react` (see https://github.com/ai/size-limit/issues/79#issuecomment-475728603).

### DIFF
--- a/packages/aragon-api-react/package.json
+++ b/packages/aragon-api-react/package.json
@@ -18,7 +18,7 @@
   },
   "scripts": {
     "build": "babel src --out-dir dist --source-maps",
-    "dev": "npm run build -- --watch"
+    "build:dev": "npm run build -- --watch"
   },
   "peerDependencies": {
     "react": "^16.8.0"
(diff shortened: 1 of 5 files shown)


packages/aragon-api-react/package.json
```diff
@@ -17,7 +17,7 @@
     "access": "public"
   },
   "scripts": {
-    "build": "babel src --out-dir dist --source-maps",
+    "build": "babel src --out-dir dist --source-maps && terser dist/index.js --mangle --mangle-props --source-map --output dist/index.js",
```
sohkai: Not sure if we _really_ need to do this; are you afraid someone will be using this without a build step themselves?

bpierre: You’re right, I was optimizing and forgot that we don’t compress our libraries: I’ll remove it.

PR #271, opened 2019-03-29
Docs update: include react api, aragon app architecture & fixes

0xGabi:
closes https://github.com/aragon/hack/issues/23

New changes to restructure aragonAPI section: https://github.com/aragon/hack/issues/102


docs/APP.md
```diff
@@ -1,4 +1,4 @@
-# App
+# aragonJS - App API
```
sohkai: Not sure if we should change the title to use aragonJS; see https://github.com/aragon/hack/pull/127#discussion_r271190596.

Maybe just "JavaScript — App API", etc. would work as headings?

packages/aragon-api-react/README.md
```diff
@@ -1,6 +1,6 @@
-# aragonAPI for React
+# aragonJS - React Hooks
```
sohkai:
```suggestion
# JavaScript - React API
```

I think using "Hooks" directly is a bit misleading, since this package is meant to be used by all React devs, not just those who want hooks.

docs/APP.md
bpierre: Shouldn’t we use “aragonAPI” instead of “App API”, since it is the main module?

We could have:

- aragonAPI − JavaScript
- aragonAPI − React
- aragonAPI − …

Or:

- aragonAPI for JavaScript
- aragonAPI for React
- aragonAPI for …

Or we could put it after, but it feels a bit weird IMO:

- JavaScript − aragonAPI
- React − aragonAPI
- … − aragonAPI

sohkai:
```
aragonAPI for JavaScript
aragonAPI for React
aragonAPI for wrappers
aragonAPI for …
```

Sounds quite nice 👍

0xGabi: I also like `aragonAPI for ...`. I'm a bit confuse still about the extent of the JavaScript implementation. I thought all `aragon.js` was the JavaScript implementation of `aragonAPI`.

docs/APP.md
```diff
@@ -1,4 +1,4 @@
-# App
+# aragonAPI for Javascript
```
0xGabi: Not sure about this name, still a bit confuse about the extent of Javascript implementation of `aragonAPI`.

2color: Perhaps aragonAPI is enough. You've already mentioned in the first page that it's the one of the reference implementations
https://github.com/aragon/hack/pull/127/files#diff-869e39af999dedb4045ca0a0c54e727aR14

docs/APP.md
2color: I understand the confusion. `aragon.js` is just the monorepo name. The suggestions above indicate the actual semantic meaning of each part of it.

docs/PROVIDERS.md
```diff
@@ -1,4 +1,4 @@
-# Providers
+# aragonAPI for providers
```
2color: These are just providers.

docs/APP.md
2color: This discussion got me thinking. Now when I think of `@aragon/wrapper` it's actually just the Client/Wrapper API in the sense that it provides an API for clients, e.g. CLI or the official client (`aragon/aragon`).

2color: Following a conversation with @0xGabi I think it's important we indicate three things:
- It's the App API
- It's part of the Javascript implementation of aragonAPI
 
Given the other page with the explanation about the JavaScript explanation I think `aragonAPI for Apps` should be fine.

docs/README.md
```diff
@@ -4,9 +4,9 @@ A JavaScript implementation of aragonAPI, used to interact with aragonOS by hand
 
 ### Docs
 
+- [React API](/packages/aragon-api-react/README.md)
```
sohkai: Similar to the other comment, I'd make this a subcategory of "App API"

docs/WRAPPER.md
```diff
@@ -1,4 +1,4 @@
-# aragonAPI for wrappers
+# aragonAPI for Wrappers
```
sohkai: Maybe we could change to say "API for Aragon client implementations"? Nobody really knows what the "wrapper" really is 😄

0xGabi: Yes, I think this is more clear

PR #273, opened 2019-04-03
Wrapper: add artifacts for aragonPM apps

izqui:
These artifacts have been generated with a new CLI command (`aragon apm extract-functions`) to extract function information from contracts that are not published as standalone Aragon apps: https://github.com/aragon/aragon-cli/pull/423.

This also allows the Aragon client to render information about aragonPM orgs (e.g. https://rinkeby.aragon.org/#/0x2df31c5fcbb94656fc98e2953cf0da03e2885cfa; our `open.aragonpm.eth` rinkeby org).


packages/aragon-wrapper/src/core/aragonOS/index.js
```diff
@@ -4,40 +4,72 @@ import { getAbi, getArtifact } from '../../interfaces'
...
+    return
+  }
+
+  return {
+    isAragonOsInternalApp: true,
+    ...appInfo
+  }
+}
+
+function getAPMAppInfo (appId) {
```
izqui: Decided to split it off in two functions so `isAragonOsInternalApp` is false for aragonPM apps. Alternatively we could add a boolean to the items in `APP_MAPPING` denoting whether the app is an internal app or not.

packages/aragon-wrapper/src/core/aragonOS/index.js
```diff
@@ -4,40 +4,72 @@ import { getAbi, getArtifact } from '../../interfaces'
...
 
+function getAragonOsInternalAppInfo (appId) {
+  const appInfo = getAppInfo(appId, 'aragon', SYSTEM_APP_MAPPINGS)
+
+  if (!appInfo) {
+    return
+  }
+
+  return {
+    isAragonOsInternalApp: true,
```
sohkai: Let's move this to be after the `...appInfo` so it's ensured to be set.

packages/aragon-wrapper/src/core/aragonOS/index.js
```diff
@@ -4,40 +4,72 @@ import { getAbi, getArtifact } from '../../interfaces'
...
+  const appInfo = getAppInfo(appId, 'apm', APM_APP_MAPPINGS)
+
+  if (!appInfo) {
+    return
+  }
+
+  return {
+    isAragonOsInternalApp: false,
+    ...appInfo
+  }
```
sohkai: I think we could simply `return getAppInfo(appId, 'apm', APM_APP_MAPPINGS)
`; `isAragonOsInternalApp` doesn't need to always be set.

packages/aragon-wrapper/src/core/aragonOS/index.js
```diff
@@ -4,40 +4,72 @@ import { getAbi, getArtifact } from '../../interfaces'
...
+}
+
 function getKernelNamespace (hash) {
   if (KERNEL_NAMESPACES.has(hash)) {
     return { name: KERNEL_NAMESPACES.get(hash), hash }
   }
 }
 
-export { getAragonOsInternalAppInfo, getKernelNamespace }
+export { getAragonOsInternalAppInfo, getAPMAppInfo, getKernelNamespace }
```
0x6431346e:
```suggestion
export { getAragonOsInternalAppInfo, getApmAppInfo, getKernelNamespace }
```
Might be worth considering to style `APM` as `Apm` in the context of a camel case name, similar to `AragonOs`

packages/aragon-wrapper/src/core/aragonOS/index.js
```diff
@@ -4,40 +4,72 @@ import { getAbi, getArtifact } from '../../interfaces'
...
+const SYSTEM_APP_MAPPINGS = new Map([
   [aragonpmAppId('acl'), 'ACL'],
   [aragonpmAppId('evmreg'), 'EVM Script Registry'],
   [aragonpmAppId('kernel'), 'Kernel']
 ])
 
+const APM_APP_MAPPINGS = new Map([
+  [aragonpmAppId('apm-registry'), 'APM Registry'],
+  [aragonpmAppId('apm-repo'), 'Repo'],
+  [aragonpmAppId('apm-enssub'), 'ENS Subdomain Registrar']
```
0x6431346e:
```suggestion
  [aragonpmAppId('apm-enssub'), 'ENS Subdomain Registrar'],
  [aragonpmAppId('apm-registry.open'), 'APM Registry'],
  [aragonpmAppId('apm-repo.open'), 'Repo'],
  [aragonpmAppId('apm-enssub.open'), 'ENS Subdomain Registrar']
```

See https://github.com/aragon/aragon-cli/pull/425#issuecomment-480488022

sohkai: Hmm, just hooked up the client to try and inspect an aragonPM instance and indeed we do need these overrides :/.

Not sure if we'll be able to do something clever but this might be the easiest way to get compatibility with the current `open.aragonpm.eth` deployment.

0x6431346e: I think this is a justifiable shortcut while we explore how to fix this properly so this contribution doesn't become stale.
I will open an issue and merge this PR with the related ones.

izqui: These repos are actually set up by `APMRegistryFactory` [here](https://github.com/aragon/aragonOS/blob/dev/contracts/factory/APMRegistryFactory.sol#L76-L94) and the appIds are always set to the `apm-*` repos in the registry. 

All registries that are created with this factory will have the same issue. The factory was developed for deploying the root registry and we didn't take this into account.

sohkai:
I've merged with master, refactoring this a little bit now that we have the `core/apm` module introduced with the `installedRepo` changes.

In summary:

- Moved most of the interface fetching to `./interfaces.js`, so that both `core/aragonOS` and `core/apm` have access to `getAppInfo()`
- Split the repo-specific utilities into `core/apm/repo`
- Unified the API to return `null` instead of `undefined` for `getApmAppInfo()`, `getAragonOsInternalAppInfo (), and `getKernelNamespace()`

cc @2color

packages/aragon-wrapper/src/interfaces.js
```diff
@@ -20,15 +27,66 @@ const ABIS = {
   'aragon/Forwarder': abiAragonForwarder.abi,
   'aragon/Kernel': abiAragonKernel.abi,
   'aragon/EVM Script Registry': abiAragonEVMScriptRegistry.abi,
-  'aragon/Repo': abiAragonRepo.abi,
+  'apm/APM Registry': abiApmRegistry.abi,
+  'apm/Repo': abiApmRepo.abi,
+  'apm/ENS Subdomain Registrar': abiApmEnsSubdomainRegistrar.abi,
```
2color: What is the canonical source for the "ABI name", e.g. `ENS Subdomain Registrar`

2color: In other words, where do the strings come from in 
https://github.com/aragon/aragon.js/blob/1ea7047a064ca3796419c89a54c1a5224dee7af6/packages/aragon-wrapper/src/interfaces.js#L52-L60

sohkai: Those are mostly made up names we chose ourselves as the "human readable app name".

They would correlate to the name in an app's `manifest.json` (e.g. [Token Manager's](https://github.com/aragon/aragon-apps/blob/master/apps/token-manager/manifest.json#L2)).

packages/aragon-wrapper/src/interfaces.js
```diff
@@ -20,15 +27,66 @@ const ABIS = {
...
 export const getArtifact = name => ARTIFACTS[name] || null
+
+export function getAppInfo (appId, namespace) {
+  const mapping = APP_NAMESPACE_MAPPINGS.get(namespace)
+
+  if (!mapping || !mapping.has(appId)) {
+    return null
+  }
+
+  const appName = mapping.get(appId)
```
2color: It seems that `appName` is in fact two different types, used in two incompatible ways. 

1. As an ENS name, e.g. `apm-registry.aragonpm.eth`. Has 1:1 relationship with the `appID`(just a hash of it)
1. As a readable case sensitive identifier, e.g. `APM Registry`

The second use case for this type is more general because a single `appName` can have multiple instances. What do you think about indicating this by renaming from `appName`?

sohkai: Good point; we generally use `name` instead for the human readable name so let's use that here too.

PR #276, opened 2019-04-05
feat: Add message signing

### LINKED ISSUE #124: Allow apps to sign arbitrary data via a RPC call
Some apps may require users to sign arbitrary messages, data, or even transactions. aragon.js should tell its server it has a signing request like this, to prompt it to show an appropriate screen (e.g. aragon/aragon's signer panel).

Schwartz10:
Fixes #124.

**Changes**


packages/aragon-wrapper/src/rpc/handlers/sign-message.js
```diff
@@ -0,0 +1,4 @@
+export default function (request, proxy, wrapper) {
+  const messageToSign = request.params[0]
+  return wrapper.signMessage({ fromAddress: proxy.address, message: messageToSign })
```
Schwartz10: wondering if this is a bit confusing. I at first thought `fromAddress` should be the address that should sign the message, not the app that sent the request.

Should we include both addresses? The front-end already has access to the `account` that will sign the message, so the wrapper doesn't _need_ to send this information along the way. Although, on the other hand, the `account` should never really differ?

packages/aragon-wrapper/src/index.js
```diff
@@ -772,6 +774,29 @@ export default class Aragon {
...
+   *
+   * @param {Array<Object>} params An object containing the address of the app and the data to be signed
+   * @return {Promise<string>} signature hash
+   */
+  signMessage (params) {
+    const { fromAddress, message } = params
+
+    return new Promise((resolve, reject) => {
+      this.signatures.next({
+        from: fromAddress,
```
Schwartz10: again, wondering if we should rename this, or also include the `account` that will sign the message

Schwartz10:
Credit to @mul1sh for help on this

Schwartz10:
Hey @sohkai this is ready for an initial review + I have a couple questions:

1. Do you think we should include the `account` that will be responsible for signing this message in the `signatureBag`? The wrapper doesn't _need_ to send this information in order to execute the signature, but `from` might feel a little misleading (besides the fact that this pattern exists for the `transactionBag` as well). Happy to keep as is and make updates down the road if needed.
2. Take a look at the change made in this commit to aragon-api to handle errors: https://github.com/aragon/aragon.js/pull/276/commits/b7e427d297a3685a6032e8f95c741de64e354002

This isn't great, but I think it's better than `pluck`ing the result, because front-ends can more easily handle error cases. However, I'm wondering if there's a better way to accomplish the same goal. Ideally, front-ends should be able to do this:

```js
aragonApi.signMessage((err, sigHash) => {
  if (err) // handle it
  // use sigHash
})
```
Because the error first callback style is familiar.

Instead, the rxjs allows a front-end dev to do this:

```js
aragonApi.signMessage(response => {
  if (response.error) // handle it
  else // use response.result for sig hash
})
```

This is way less ideal because it's unfamiliar. 

I'm an rxjs noob, so is there a better, simple way to imitate the normal callback style using observables?

Schwartz10:
Once we settle on the API interface from the comment above, I'll update the docs

Schwartz10:
Linking https://github.com/aragonone/product/issues/33

packages/aragon-wrapper/src/index.js
```diff
@@ -772,6 +774,29 @@ export default class Aragon {
...
   }
 
+  /**
+   * Allows apps to sign arbitrary data via a RPC call
+   *
+   * @param {Array<Object>} params An object containing the address of the app and the message to be signed
+   * @return {Promise<string>} signature hash
+   */
+  signMessage (params) {
+    const { fromAddress, message } = params
```
sohkai: I'd just have `signMessage(message)` be the function signature

packages/aragon-api/src/index.js
```diff
@@ -332,6 +332,18 @@ export class AppProxy {
...
     )
   }
+
+  /**
+   * Allow apps to sign arbitrary data via a RPC call
+   *
+   * @param  {string} msgToSign
+   * @return {void}
+   */
+  signMessage (msgToSign) {
```
sohkai: I'd rename this method to `requestSignMessage()`, since the app ultimately doesn't control if the message gets signed or not.

sohkai:
> 1. Do you think we should include the account that will be responsible for signing this message in the signatureBag?

No, this shouldn't be necessary. An aragon client should be responsible for connecting to an Ethereum provider that provides the connected account that will sign the message.

> This pattern exists for the `transactionBag` as well

The `transactionBag` is purely an aragon/aragon concept, but I'm not sure I understand what you mean.

`from` should _always_ be the sender address of a transaction. We do a bit more processing in [`getTransactionPath()`](https://github.com/aragon/aragon.js/blob/master/packages/aragon-wrapper/src/index.js#L879) against all declared accounts to help users find a transaction path if they've connected more than one account with aragon.js (this never happens in aragon/aragon).

> 2. Take a look at the change made in this commit to aragon-api to handle errors

The promise returned from `@aragon/wrapper`'s `signMessage()` can be rejected, which will end up rejecting on the `@aragon/api` caller's side as well. This is how an app would detect that their sign request was cancelled or had failed.

Schwartz10:
Hi @sohkai thank you for reviewing. I'm going to address all your comments (including the two inline ones) here. I want to make sure I'm following the desired logic properly.

aragon/aragon needs 3 pieces of information to execute a signature and follow the given [designs](https://github.com/aragon/aragon/pull/488#issuecomment-476213816):

1. the message to sign
2. the account to sign the message from 
3. the app that requested the signature

(1) is easy to handle, but there are different options for how we handle (2) and (3) with respect to _getting_ and _passing_ the data (between packages), so I'm going to go through both and make sure we're on the same page.

**(2) - the account to sign the message from**

> An aragon client should be responsible for connecting to an Ethereum provider that provides the connected account that will sign the message. 

I see three ways to achieve this:

 a. The aragon client app developer could pass the account to sign from as a second parameter in the call to `aragon/api.requestSignMessage` like: `app.requestSignMessage(msg, account)`.
b. Once the call to `aragon/api.requestSignMessage` reaches `aragon/wrapper`, we could get the account to sign from through `this.accounts`. This would mean the call to `app.requestSignMessage` would not specify the `account` to sign from. 
c. Once the call reaches aragon/aragon, we could use the `walletWeb3` to get the current account (this also means that the call to `app.requestSignMessage` would not specify the `account` to sign from)

Which option do you think makes the most sense? 

**(3) - the app that requested the message signature**

> I'd just have signMessage(message) be the function signature

If I'm understanding correctly, you can access the app address that made the call to aragon/api through `proxy.address` [here](https://github.com/aragon/aragon.js/pull/276/files/93233a8cfd4b4e5135b1a6f821bfc478824cbf9c#diff-dc86c3a663278d85524b278aab146994R1). Is this is the best place to grab the requesting app's address? If so, doesn't the `signMessage` signature need to include another param? If not, where is the best place to grab the requesting app address?

Lastly, about _documentation_:

> The promise returned from @aragon/wrapper's signMessage() can be rejected, which should end up rejecting on the @aragon/api caller's side as well. This is how an app would detect that their sign request was cancelled or had failed.

So does this mean that the documentation should suggest promisifying the Observable returned by `api.requestSignMessage`, and wrapping it in a try catch (or using `.then`, `.catch`)?

sohkai:
> (2) - the account to sign the message from

Definitely c. This is something the client should handle itself, since it should own the connection to your "wallet" Ethereum provider.

The other two options would essentially be the same thing, since the same account is injected into `@aragon/wrapper` (and `@aragon/api` just pulls this account), but would introduce unnecessary clutter into the API.

> (3) - the app that requested the message signature
> Is this is the best place to grab the requesting app's address?

Yes, the handler is the best place to grab this address. There shouldn't be a need to include another parameter in `requestSignMessage()` on `@aragon/api`'s side since all of this information is available in `@aragon/wrapper`.

> So does this mean that the documentation should suggest promisifying the Observable returned by api.requestSignMessage, and wrapping it in a try catch (or using .then, .catch)?

There is technically some other development work that needs to be done too, to get this promise accept / reject. You can basically duplicate the pattern [here](https://github.com/aragon/aragon.js/blob/master/packages/aragon-wrapper/src/index.js#L781) or [here](https://github.com/aragon/aragon.js/blob/master/packages/aragon-wrapper/src/index.js#L514), where the emitted object just holds an `resolve` and `reject` key that will be handled by the client.

The propagation of this error has been implemented in https://github.com/aragon/aragon.js/pull/277.

As for the documentation itself, you can use some of the wording from https://github.com/aragon/aragon.js/blob/master/docs/APP.md#requestaddressidentitymodification 😄.

Schwartz10:
> Definitely c.

Understood.

> Yes, the handler is the best place to grab this address. There shouldn't be a need to include another parameter in requestSignMessage() on @aragon/api's side since all of this information is available in @aragon/wrapper.

Also understood. `requestSignMessage` (in `aragon/api`) does only take one param. However, your comment [here](https://github.com/aragon/aragon.js/pull/276#discussion_r273003950) says we should only need one parameter on the `aragon/wrapper` method. I think we're on the same page here, so take a look at the current code and let me know what you think. 

> The propagation of this error has been implemented in

Great, when i pulled in #277 this is working well. Although I think for documentation purposes it makes sense to suggest promisifying the `requestSignMessage` because you can handle the errors easily in a `catch` block (unless there's a good way to handle errors in a `subscribe` operator?). I updated the docs, so you can let me know what you think.

--
If the changes are good, I think it's ready to be pulled in!

packages/aragon-wrapper/src/index.js
```diff
@@ -772,6 +774,29 @@ export default class Aragon {
     return this.accounts.pipe(first()).toPromise()
   }
 
+  /**
+   * Allows apps to sign arbitrary data via a RPC call
+   *
+   * @param {Array<Object>} params An object containing the address of the app and the message to be signed
+   * @return {Promise<string>} signature hash
+   */
+  signMessage (params) {
```
sohkai: >  However, your comment here says we should only need one parameter on the aragon/wrapper method

Ahh, yeah, actually I think it's better to go back to the two-param method instead (`signMessage(message, fromApp)`). Hadn't considered that this required the originating app when I first reviewed.

Sorry for the churn 🙏!

sohkai:
> unless there's a good way to handle errors in a subscribe operator?

Yes, [it's the second callback to `.subscribe()`](https://blog.angular-university.io/rxjs-error-handling/) 😄 .

packages/aragon-wrapper/src/index.js
Schwartz10: No worries! Is the variable name `appAddress` ok? Or would you like `fromApp`? Wondering bc i'll need to make the updates in https://github.com/aragon/aragon/pull/683 as well

Schwartz10:
> Yes, it's the second callback to .subscribe() 😄 .

Ha, realized I had tried the second error parameter before #277 was merged. Oops, brain fart. 

I'll update documentation too

packages/aragon-wrapper/src/index.js
Schwartz10: I kept as `appAddress` for now, but if you would like me to change it that's no problem.

sohkai: Actually, maybe `requestingApp`?

Just trying to find something descriptive 😄

Schwartz10: hmmm yeah. I chose `appAddress` because it's used by the permissions manager in aragon/wrapper

Schwartz10:
@sohkai changed variable name to `requestingApp`. Hopefully this is all set now!

sohkai:
Looking good, thanks @Schwartz10!

packages/aragon-wrapper/src/index.js
```diff
@@ -772,6 +774,27 @@ export default class Aragon {
     return this.accounts.pipe(first()).toPromise()
   }
 
+  /**
+   * Allows apps to sign arbitrary data via a RPC call
+   *
+   * @param {Array<Object>} params An object containing the address of the app and the message to be signed
```
```suggestion
   * @param {...} message ....
```

packages/aragon-wrapper/src/index.js
```diff
@@ -772,6 +774,27 @@ export default class Aragon {
...
+   * @return {Promise<string>} signature hash
+   */
+  signMessage (message, requestingApp) {
+    return new Promise((resolve, reject) => {
+      this.signatures.next({
+        requestingApp,
+        message,
+        accept (signatureHash) {
```
2color: @sohkai I thought the long term goal was to rename to `resolve` as we did with [identity](https://github.com/aragon/aragon.js/blob/31701f5e183ce8e5dca4785da1d2843bb75f5297/packages/aragon-wrapper/src/index.js#L519) This is not critical for this PR but should be consistent. 

Any preferences?

sohkai: Ah yes, good catch. Let's rename this to `resolve()` 👍.

```suggestion
        resolve (signatureHash) {
```

Schwartz10:
Hey thanks for wrapping this up @2color and @sohkai ! Very much appreciated.

PR #277, opened 2019-04-08
API: propagate errors to single-response APIs

sohkai:
Now callers of single response APIs will be able to catch errors from the call that are propagated through `@aragon/wrapper`.

They can either do this by including an error handler to their `.subscribe()` or by turning the returned observable into a promise and using the standard promise-rejection handling methods.

A good example of where this is useful is for transaction signing and message signing, where the user may ultimately reject or cancel the intent.

### DIFF
--- a/packages/aragon-rpc-messenger/src/index.js
+++ b/packages/aragon-rpc-messenger/src/index.js
@@ -2,7 +2,7 @@ import jsonrpc from './jsonrpc'
 import MessagePortMessage from './providers/MessagePortMessage'
 import WindowMessage from './providers/WindowMessage'
 import DevMessage from './providers/DevMessage'
-import { first, filter } from 'rxjs/operators'
+import { first, filter, map } from 'rxjs/operators'
 
 export const providers = {
   MessagePortMessage,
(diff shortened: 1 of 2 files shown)


packages/aragon-rpc-messenger/src/index.js
```diff
@@ -91,7 +91,14 @@ export default class Messenger {
     const id = this.send(method, params)
 
     return this.responses().pipe(
-      filter((message) => message.id === id)
+      filter((message) => message.id === id),
+      map((response) => {
+        if (response.error) {
```
2color: Is there any reason not to leave the `response.error` an error instance in https://github.com/aragon/aragon.js/blob/master/packages/aragon-rpc-messenger/src/jsonrpc.js#L19?

It seems like we are converting an error to a string in jsonrpc.js and back to an error here.

sohkai: Primarily because sending an `Error` instance through `postMessage()` will fail: [see things that don't work with the structured clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#Things_that_don't_work_with_structured_clone).

PR #282, opened 2019-04-16
enforce typeof message to sign is string

Schwartz10:


packages/aragon-wrapper/src/rpc/handlers/sign-message.js
```diff
@@ -1,4 +1,7 @@
 export default function (request, proxy, wrapper) {
   const message = request.params[0]
+  if (typeof message !== 'string') {
+    throw new Error('typeof message to sign must be a string')
```
sohkai:
```suggestion
    throw new Error('Message to sign must be a string')
```

sohkai:
Moved the error handling logic to `signMessage()`, since it's exposed to `@aragon/wrapper` consumers, and added some tests for the observable :).

PR #285, opened 2019-04-17
fix: avoid infinitely looping through forwarders when looking for a transaction path

### LINKED ISSUE #235: Wrapper: optimize transaction pathing for common cases
We will [soon make it a default that only token holders can create votes](https://github.com/aragon/dao-kits/pull/50) (forcing more users to use transaction pathing for simple actions).

Right now the transaction pathing algorithm either gets stuck, or takes _incredibly_ long to find these paths (or conclude no path exists) by its brute-force algorithm. We should use some of the permission information we receive through the ACL to make this quicker; if an action is only allowed by a particular subset of apps or addresses, we can simply check if the user's current address can forward through those apps and if not, conclude that no path exists.

sohkai:
Fixes #235.

Turns out the transaction pathing algorithm is sufficiently smart, we just weren't handling the 0 case correctly (resulting in an infinite loop).

The reason we can't rely on permissions data once we start looking for forwarding path is because those aren't determined by permissions; we have to check `canForward()` for each step.

An additional optimization we could make is cache which forwarders can be linked together, but this seems premature since there's unlikely to be very large paths at this point.

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -1390,7 +1390,6 @@ export default class Aragon {
         return [directTransaction]
       }
 
-      // TODO: Support multiple needed roles?
       const roleSig = app.roles.find(
         (role) => role.id === method.roles[0]
       ).bytes
@@ -1402,7 +1401,7 @@ export default class Aragon {
         []
       )
 
-      // No one has access
+      // No one has access, so of course we (or the final forwarder) don't as well
       if (appsWithPermissionForMethod.length === 0) {
         return []
       }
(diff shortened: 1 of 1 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -1507,19 +1506,21 @@ export default class Aragon {
...
+    // We eagerly evaluate paths that look promising; as soon as we find a forwarder that
+    // "links" up with a previous forwarder, we continue exploring this path immediately.
     // TODO(onbjerg): Should we find and return multiple paths?
     do {
       const [path, [forwarder, ...nextQueue]] = queue.shift()
 
-      // Skip paths longer than 10
-      if (path.length > 10) continue
+      // Skip if no forwarder or the path is longer than 5
+      if (!forwarder || path.length > 5) continue
```
sohkai: The `!forwarder` check was the important one we were missing before; even though `nextQueue` would be empty, we'd still treat it as a normal path and re-add it back to the path at the end.

PR #286, opened 2019-04-17
wrapper: handle intent baskets

sohkai:
Has a few kinks to work out, but works for single-transaction organization upgrades.

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -1246,7 +1246,9 @@ export default class Aragon {
       )
 
       if (path.length > 0) {
-        return this.describeTransactionPath(path)
+        try {
+          return this.describeTransactionPath(path)
+        } catch (_) { }
       }
     }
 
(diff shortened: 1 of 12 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -1242,6 +1246,99 @@ export default class Aragon {
...
+        async ([destination, methodName, params]) =>
+          createDirectTransaction(sender, await this.getApp(destination), methodName, params, this.web3)
+      )
+    )
+
+    if (intentPaths[0].length === 1) {
+      // Sender has direct access
+      return {
+        direct: true,
+        transactions: directTransactions
```
izqui: Don't we need to `applyTransactionGas` to these transactions before returning them?

sohkai: Hmm yes... we should also apply gas to https://github.com/aragon/aragon.js/blob/ca9f562a4b14783ced5087e2aae8217453e2e8cf/packages/aragon-wrapper/src/index.js#L1776.

In the case where apply the gas fails, I guess we should just return an empty array to signal that the transaction wouldn't succeed.

packages/aragon-wrapper/src/utils/transactions.js
```diff
@@ -0,0 +1,110 @@
+import { toBN } from 'web3-utils'
+import { getAbi } from '../interfaces'
+
+const DEFAULT_GAS_FUZZ_FACTOR = 1.5
+const PREVIOUS_BLOCK_GAS_LIMIT_FACTOR = 0.95
```
2color: How did you arrive at this number? is it reduced as a safety margin?

sohkai: More or less just picked one. The actual possible decrease is only `previous limit / 1024`, which is quite small, but we generally want at least a small safety margin over the block gas limit anyway.

packages/aragon-wrapper/src/utils/index.js
```diff
@@ -51,20 +49,13 @@ export function makeProxyFromABI (address, abi, web3, initializationBlock) {
...
-  const bufferedGasLimit = Math.round(estimatedGasLimit * gasFuzzFactor)
-
-  if (estimatedGasLimit > upperGasLimit) {
-    // TODO: Consider whether we should throw an error rather than returning with a high gas limit
-    return estimatedGasLimit
-  } else if (bufferedGasLimit < upperGasLimit) {
-    return bufferedGasLimit
-  } else {
-    return upperGasLimit
+export function findMethodOnAppFromData (data, app) {
```
2color: What type does data have? Can this function be documented?

packages/aragon-wrapper/src/index.js
```diff
@@ -1242,6 +1246,99 @@ export default class Aragon {
...
+   *
+   * @param  {Array<Array<string, string, Array<*>>>} intentBasket Intents
+   * @param  {Object} [options]
+   * @param  {string} [options.checkMode] Path checking mode
+   * @return {Promise<Array<Object>>} An array of Ethereum transactions that describe each step in the path
+   */
+  async getTransactionPathForIntentBasket (intentBasket, { checkMode = 'all' } = {}) {
+    // Get transaction paths for entire basket
+    const intentsToCheck =
+      checkMode === 'all'
```
2color: When would checkMode `all` be used? 

The only usage of it uses `single`:
https://github.com/aragon/aragon/blob/98bf8de6309ad747c7cea1c92e4ffe08823e296c/src/components/Upgrade/UpgradeOrganizationPanel.js#L69-L72

2color: Is that because the intentBasket coming from the client will have the same path?

2color:
I tested the direct transaction scenario on an org with an account that has the `Manage apps` permission and it only upgraded the voting app. See the [transaction](https://rinkeby.etherscan.io/tx/0xc15d40a46747579d46037a6008540580d55ce45c4df165e36628f394a5c9b508) and the [org](http://localhost:3000/#/3color.aragonid.eth/apps?p=installed)

sohkai:
> I tested the direct transaction scenario on an org with an account that has the Manage apps permission and it only upgraded the voting app.

Yes... debugging this with the client now. The problem is that if you have direct access, the only thing we can do is ask you to sign multiple transactions (and this bit isn't working well yet).

packages/aragon-wrapper/src/index.js
sohkai: `all` isn't used yet, since our only use case is calling the same function, but in the future when you might be including _very different_ transactions (e.g. withdraw from Finance and burn a token at the same time, see https://github.com/aragon/aragon/issues/363), it'll be important to use `all` to check that all the actions can actually be performed together.

packages/aragon-wrapper/src/utils/index.js
sohkai: `data` is the calldata ([`function selector`](https://solidity.readthedocs.io/en/v0.5.7/abi-spec.html#function-selector) + encoded arguments)

packages/aragon-wrapper/src/index.js
```diff
@@ -1242,6 +1246,99 @@ export default class Aragon {
...
+          ? [intentBasket[0]] // single -- only use first intent
+          : []
+    const intentPaths = await Promise.all(
+      intentsToCheck.map(
+        ([destination, methodName, params]) =>
+          this.getTransactionPath(destination, methodName, params))
+    )
+
+    // Check paths all match
+    const individualPaths = intentPaths
```
2color: What do you think about breaking this into a separate utility function to check for path uniqueness?

packages/aragon-wrapper/src/index.js
```diff
@@ -1242,6 +1246,99 @@ export default class Aragon {
...
+   *
+   * Allows user to specify how many of the intents should be checked.
+   * `checkMode` supports:
+   *   'all': All intents will be checked to make sure they use the same forwarding path.
+   *   'single': assumes all intents can use the path found from the first intent
+   *
+   * @param  {Array<Array<string, string, Array<*>>>} intentBasket Intents
+   * @param  {Object} [options]
+   * @param  {string} [options.checkMode] Path checking mode
+   * @return {Promise<Array<Object>>} An array of Ethereum transactions that describe each step in the path
```
2color: The return type here is an object of two types depeneding on whether it's direct or not.

packages/aragon-wrapper/src/index.js
2color: What do you think about breaking this into separate functions?

packages/aragon-wrapper/src/index.js
```diff
@@ -1456,6 +1549,33 @@ export default class Aragon {
...
+    const allKernel = intents.every(({ to }) => this.isKernelAddress(to))
+    if (!allKernel) return
+
+    const kernelApp = await this.getApp(intents[0].to)
+    const allSetApp = intents.every(({ data }) => {
+      const intentMethod = findMethodOnAppFromData(data, kernelApp)
+      return intentMethod && intentMethod.sig === 'setApp(bytes32,bytes32,address)'
+    })
+
+    // Just assume for now updating all four apps (finance, token manager, vault, voting)
```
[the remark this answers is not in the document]
sohkai: This has now been changed to check the upgrades against a list of "known" app Ids.

sohkai:
The logic for how radspec is handled as been completely refactored into its own module. In the future we may think about letting users of `@aragon/wrapper` define their own radspec "preprocessors".

This also strips out a few unnecessary details out of the intents:

- `identifier`: this was broken as part of `5.0.0-beta.1` anyway, since the identifier information has now been split from the apps observable
- `name`: this is easily found by users if they match a path's `to` address with the corresponding app object

2color:
Still getting an error:
Screen Shot 2019-04-30 at 10 07 31 am

sohkai:
@2color Are you using the current head of https://github.com/aragon/aragon/pull/732? It should be handling the removal of `name` now.

PR #295, opened 2019-04-30
wrapper: rename entrypoint to help sourcemaps

2color:
## What

- Rename entrypoint file to @aragon/wrapper

## Why

- For some strange reason Chrome and Firefox don't seem to be able to parse the source map properly when debugging `aragon/aragon` with `@aragon/wrapper` linked. 
- This extends to our production deployments took, which are deployed with the sourcemaps. Yet for some strange reason, this module isn't picked up properly. See in screenshot below how the module is missing.
Screen Shot 2019-04-30 at 10 25 29 am


2color:
@bpierre 
Thanks for pointing this out. We should apply this change to the other packages. 

I'm curious what the root cause of this bug is. 

For example, our production build which includes the sourcemap seems to mix files from the `aragon/aragon` repository and `@aragon/wrapper` package. See below the `cache/index.js` and the `apps` folder, each of which are coming from different repos.

Screen Shot 2019-05-06 at 12 28 32 pm

sohkai:
I think the Chrome debugger is likely confused with the source map paths, since we export them transpiled in the aragon.js repo. I imagine it's having a hard time knowing that it's in `node_modules`, when the source map probably says everything is from `../src/index.js` :shrug:.

packages/aragon-wrapper/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@aragon/wrapper",
   "version": "5.0.0-rc.3",
   "description": "Library for Aragon client implementations",
-  "main": "dist/index.js",
+  "main": "dist/aragon-wrapper.js",
```
sohkai: What do you think about keeping an `index.js` file that just re-exports the other modules, to be more similar to other npm packages?

We also do some re-exporting in the current `index.js` that could be removed from `aragon-wrapper.js`.

sohkai:
And on `size-limit`, we should probably specify a glob instead:

```
"path": "dist/*.js"
```

packages/aragon-wrapper/package.json
2color: I like the idea. I also had mixed feelings about this given all the other npm packages I've seen.

2color:
> And on size-limit, we should probably specify a glob instead:
> "path": "dist/*.js"

I'm not sure what you're actually suggesting. size-limit expects an entry point and for every entrypoint we have a size limit so I'm not sure what we'll get with a glob?

sohkai:
Ahh, didn't realize `size-limit` was smart enough to compute all the dependencies from a single entry point :+1:.

PR #296, opened 2019-04-30
Add eslint and prettier to packages

2color:
## What

- Add eslint and prettier configuration based on https://github.com/aragon/aragon 
- Add required dependencies to packages
- Replace `standard` with `eslint` in lint scripts

## Why
- It's prettier

## TODO
- [ ] Agree on the rules with the rest of the team
- [ ] `prettier --write`
- [ ] Add to `react-api`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,29 @@
+{
+  "env": {
+    "browser": true,
+    "es6": true
+  },
+  "extends": [
+    "plugin:import/recommended",
+    "plugin:promise/recommended",
+    "standard",
+    "standard-react",
+    "prettier/react",
+    "plugin:prettier/recommended"
+  ],
+  "parser": "babel-eslint",
+  "parserOptions": {
+    "ecmaFeatures": {
+      "experimentalObjectRestSpread": true
+    },
+    "sourceType": "module"
+  },
+  "plugins": ["prettier", "react", "import", "promise"],
+  "rules": {
+    "import/no-unresolved": ["error", { ignore: ["^react(-dom)?$", "^styled-components$"] }],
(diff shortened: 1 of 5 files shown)


AquiGorka:
@2color great! 🎉 

Maybe this is a good place to continue with the styleguide splike (https://github.com/aragonone/product/issues/62)

2color:
@AquiGorka Can you share more info on what you already did?

AquiGorka:
Created a repo [`javascript-stylguide`](https://github.com/aragonone/aragon-javascript-styleguide) where rules and configs are kept (versions are set here for everyone).
Started implementing this for [`apps`](https://github.com/aragon/aragon-apps/pull/567).

There are differences between the versions of prettier between some apps so I had to take a lot of time to update the format to a version all apps accept and then start updating versions.

2color:
> Created a repo javascript-stylguide where rules and configs are kept (versions are set here for everyone) and then started implementing it for apps.

Great! Can we make it public and publish the package?

AquiGorka:
> Great! Can we make it public and publish the package?

Of course, it was originally public and then moved to the `aragonone` organization, what would be the approach here @sohkai ? Do we move it back and open source here or make it public under the `aragonone` org?

sohkai:
@AquiGorka @2color We should move it back under the Aragon organization, since it'll be applied to repos under this organization as well.

Let's leave this in limbo for a bit—I'd like to restructure the in-progress styleguide repo so it also has configurations for babel, react / vanilla JS / node specific environments, along with a separate Solidity section.

The main reason why `aragon-apps` is a PITA is because we didn't complete the `babel@7` upgrade there, but this should be smoothed over with a shared config.

Issue #294, opened 2019-04-29
State caching - app state


2color:
## Background

When interacting with the Aragon client, there are a couple of factors that make the initial load very slow. This is primarily because we fetch logs/events from a fixed point (a certain block number) in the past. 

The initial load of the voting app for the [Aragon Network Vote #2](https://mainnet.aragon.org/#/governance.aragonproject.eth/0x277bfcf7c2e162cb1ac3e9ae228a3132a75f83d4) takes around 2 minutes in the best case. 

Currently, the reduced state used for rendering is cached but all past events are fetched from the beginning. That means that while the second load will render fast, if any changes – like a vote cast – happened between the first load and second load, it will take around 2 minutes for them to be visible to the user.

## Goal
- Eliminate the need to refetch state that is unlikely to change
- Take chain reorgs into account 
  - Here we can assume that after a certain threshold, a block isn't likely to change
- Provide the necessary states to render the different loading states https://github.com/aragonone/product/issues/27 

### Currently
- Currently, apps have two kinds of state:
  - event state - state derived from streamed events
    - Unlikely to change and can be considered safe to cache
  - contract state - state derived by making calls to the contract.
- Both state types can come from the app's contract or an external contract, e.g. the [token contract in the voting app](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/script.js#L64). 

### Next steps 
- Update aragonAPI's `store` interface to take a new init function for loading contract state (state fetched directly from the blockchain without event reduction).
This should replicate the logic of the `INITIALIZATION_TRIGGER` and give a consistent interface to apps fetching contract state. `app.state(reducer, events:, init: () => Promise<object> )`
- Add external contract subscription mechanism to aragonAPI's `store` interface (potentially replace the current `events` param). Allow passing external contracts (`address`, `abi`), for which the aragon api will handle fetching and subscribing to events. 
  - Still need to investigate what how this API should look and if it has any benefits over the current `events` array that can be passed into the `store` function. 
- Separate aragonAPI's store `web3.contract.events` calls into two calls to `getPastEvents` and `events`
  - By separating the two calls we can guarantee we have the correct state for caching at a specific block number when all (given external contracts for which events are fetched) calls to  `getPastEvents` have completed. 
- Cache the reduced state conditionally when it's old enough to be considered safe and unlikely to reorg.

## Potential challenges

[...]

Issue #206, opened 2018-11-12
Apps: add ability for apps to know if they are in the inital loading phase


bpierre:
Could be related to #86 if we want to have the same or a similar API to get these two pieces of information.

sohkai:
Something we could also do is detect if we're syncing a _large_ amount of blocks (e.g. >100k) and then successively call `eth_getlogs` with `to` and `from` block numbres so that we don't have to wait for the entire operation to finish at once before we process any results.

As of today, loading the `governance.aragonproject.eth` org's voting app requires >5min to sync ~1million blocks, with the `eth_getlogs` sending 20kb of data all at once at the end. We could instead split this operation into 100k block chunks to start loading data faster for the user.

sohkai:
Closing since #297 implements this functionality, and created a new issue for the incremental sync (#331).

PR #297, opened 2019-04-30
App state caching

2color:
## What
- Split app state event streams into two separate streams for past and future events
- Add an `init` option for initialising the state (to replace the `INITIALIZATION_TRIGGER` pattern)
- Deprecate the `events` option in favour of an `externals` option for external contracts

## TODO
- [x] Accounts trigger built into the store
- [x] Provide a mechanism for caching external events. For this, app scripts should be able to pass in an external contract ([as here](https://github.com/aragon/aragon-apps/blob/d49276aa5d79030ca4d59d8f74197ec97459e5a1/apps/finance/app/src/script.js#L161)) and pass a `fromBlock` used for fetching events the events. 
- [x] Differentiate between actual cache and state passing to application (which uses the `state` cache key)
- [x] Pass `cachedState` into `init` function - https://github.com/aragon/aragon.js/pull/297#discussion_r285333612
- [x] emit a custom event for the reducer to know the sync status (the `endWith` operator might be useful)
- [x] Replace the external `pastEvents` handler's use of the `Proxy` class with `web3.eth.Contract`

- [x] Add web3.js options support to external's and `pastEvents` API
- [x] Add web3.js options support to external's `events` API (#310)

## TODO separately 
- Add `syncStatus())` for apps to know the sync status -https://github.com/aragon/aragon.js/pull/297#discussion_r285333605 

## Corresponding PRs in apps to use updated API
Voting - https://github.com/aragon/aragon-apps/pull/836
Token Manager - https://github.com/aragon/aragon-apps/pull/845
Finance - https://github.com/aragon/aragon-apps/pull/854

## Related issues
Fixes :
 - #294
 - #206
 -  #6 
 -  #63 
 - https://github.com/aragonone/product/issues/98 - by introduces the sync states
- Supersedes https://github.com/aragon/aragon.js/pull/85.


packages/aragon-wrapper/src/rpc/handlers/past-events.js
```diff
@@ -0,0 +1,14 @@
...
+  if (fromBlock) {
+    eventsOptions.fromBlock = fromBlock
+  }
+
+  if (toBlock) {
+    eventsOptions.toBlock = toBlock
+  }
+
+  return proxy.pastEvents(null, eventsOptions)
+}
```
sohkai: What do you think about abstracting this all into the `Proxy.events()`, so any user of the Proxy will get these benefits? We can emit a special sentinel "event" to notify the user that they've gone from `eth_getlogs` to `eth_subscribe`.

packages/aragon-wrapper/src/core/proxy/index.js
```diff
@@ -12,6 +12,38 @@ export default class Proxy {
...
+    }
+
+    if (eventNames.length === 1) {
+      //
+      return from(
+        this.contract.getPastEvents(eventNames[0], { fromBlock, toBlock })
+      )
+    } else {
+      // Get all events in the block range and filter
+      return from(
```
sohkai: Interesting, does `fromEvent()` not work here? I would've expected it to be completely the same as `.events()`.

packages/aragon-api/src/index.js
```diff
@@ -214,10 +251,16 @@ export class AppProxy {
    *
    * @param  {Function} reducer A function that reduces events to a state. This can return a Promise that resolves to a new state.
    * @param  {Observable[]} [events] An optional array of `Observable`s to merge in with the internal events observable
+   * @param  {Function} [init] An   optional array of `Observable`s to merge in with the internal events observable
    * @return {Observable} An [RxJS observable](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) that emits the application state every time it changes. The type of the emitted values is application specific.
    */
-  store (reducer, events = [empty()]) {
-    const initialState = this.state().pipe(first())
+  store (reducer, events = [empty()], init) {
```
sohkai: I like that this is kind of backwards compatible, but I think it'd be much more clear if it were something like:

```js
store (init, reducer, { externalEvents })
```

packages/aragon-api/src/index.js
```diff
@@ -231,16 +274,49 @@ export class AppProxy {
         Promise.resolve(reducer(state, event))
```
sohkai: I think with RxJS 6 we can now remove the `Promise.resolve()` and just use `from()` directly :smile:

packages/aragon-api/src/index.js
```diff
@@ -231,16 +274,49 @@ export class AppProxy {
...
+          tap((state) => {
+            console.count('cache pastState')
+            this.cache(CACHED_BLOCK_KEY, pastEventsToBlock)
+            this.cache('state', state)
+          })
+        )
+        const currentState$ = pastState$.pipe(
+          // Use the last past state as the initial state for reducing current/future states
+          last(),
+          switchMap(pastState => merge(currentEvents$, ...events).pipe(
```
sohkai: Something to be careful of here is to synchronize the external `events`'s cached block height with the app contract's cached block height (so we'd need to do the `pastEvents` for each of these external events).

packages/aragon-wrapper/src/core/proxy/index.js
2color: `getPastEvents` only returns a promise so fromEvent doesn't work. 

https://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#getpastevents

packages/aragon-api/src/index.js
2color: That could work

packages/aragon-api/src/index.js
2color: As discussed. This is to ensure `from` doesn't mistakenly return an observable with multiple emissions in cases where a reducer returns an array.

packages/aragon-wrapper/src/rpc/handlers/past-events.js
sohkai: Discussed this offline and decided to keep the split between the `pastEvents()` and `events()`:

- More similar to native web3.js API (`getPastEvents()` and `events()`)
- Including this inside the `@aragon/wrapper`'s `Proxy` is not immediately useful for that package
- Scopes these changes to aragonAPI, where it's most useful

packages/aragon-api/src/index.js
```diff
@@ -228,19 +308,74 @@ export class AppProxy {
...
+        console.debug(`- store - currentEvents$: from: ${pastEventsToBlock} -> future`)
+        // currentEvents$ is an observable for events starting from pastEventsToBlock
+        const currentEvents$ = getCurrentEvents(pastEventsToBlock)
+
+        const pastState$ = getPastEvents(cachedBlock, pastEventsToBlock).pipe(
+          // Prevent multiple subscriptions invoking duplicate calls
+          share(),
+          // single emission array of all pastEvents -> flatten to process events
+          flatMap(pastEvents => from(pastEvents)),
+          mergeScan(wrappedReducer, { ...cachedState, ...initState }, 1),
```
2color: This could be a place for further potential performance improvements on the first load.
The rough idea would be to change assumptions to handle side effects concurrently instead of blocking on every event's side effects.

A natural example is the vault which loads blocks to update the balances on every event. Now that the fetching of past events is a finite process, we can load all balances after reducing all past events. This would require another API change, perhaps in the form of an async callback that gets the state after reducing all past event.

packages/aragon-api/src/index.js
```diff
@@ -228,19 +308,74 @@ export class AppProxy {
...
-          mergeScan(wrappedReducer, initialState, 1),
-          map((state) => this.cache('state', state))
+    const getCurrentEvents = (fromBlock) => merge(
+      this.events(fromBlock),
+      ...externals.map(({ contract }) => contract.events(fromBlock))
+    )
+
+    // If `cachedFromBlock` is null there's no cache, `pastEvents` will use the initialisationBlock
+    // External contracts can specify their own `initializationBlock` which will be used in case the cache is empty
+    const getPastEvents = (cachedFromBlock, toBlock) => merge(
```
2color: If we want to introduce a loading state for the fetching of past events, we probably want to use `concat` instead of `merge` and add a trigger/custom event before and after.

sohkai: We can use [`endWith()`](https://rxjs-dev.firebaseapp.com/api/operators/endWith) in this observable to emit this custom event.

`merge()` is nice so the reducer can start processing any of the contract events as they arrive.

---------

Alternatively (or in combination, both the reducer and an app's frontend may be interested in sync states), we may also want to use a separate API stream (e.g. `syncStatus()`) for this to not pollute the state cache. This would be really useful later on, when we fetch past blocks incrementally and could notify subscribers that we're past a certain stage, e.g.

```js
{
  status: 'SYNCING',
  details: {
    current: '<block number>',
    from: '<block number>',
    target: '<block number>',
  }
}

{
  status: 'SYNCED',
  details: {}
}
```

packages/aragon-api/src/index.js
sohkai: This is true, but I think it's already been unlocked with the `initialization` function, and doing this work immediately on initialization is actually better as it'd be much sooner than after the past events.

Here's a sketch for the vault:

1. Initialize: reload all balances of known tokens
1. Past events: in the `reducer()`, until it sees the special load trigger, avoid refreshing balances of already known tokens. If there are new tokens, fetch them
1. New events: refetch all required state in the `reducer()`

2color: True. This requires a small api change to pass the `cachedState` into the `init` function so that it can reload all balances for known tokens.

packages/aragon-wrapper/src/rpc/handlers/external.js
```diff
@@ -37,3 +38,26 @@ export function events (request, proxy, wrapper) {
...
+  const eventsOptions = {}
+  const [
+    address,
+    jsonInterface,
+    providedFromBlock,
+    providedToBlock
+  ] = request.params
+
+  // Use the app proxy's initialization block by default
+  const externalProxy = new ContractProxy(address, jsonInterface, wrapper.web3, proxy.initializationBlock)
```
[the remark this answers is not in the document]
2color: Agreed. I had the same discomfort about using it here, in part also due to `events` not using it.

packages/aragon-api/src/index.js
```diff
@@ -185,6 +229,21 @@ export class AppProxy {
...
+   * @return {Observable} An [RxJS observable](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) that emits the cached value every time it changes. The type of the emitted values is cache key specific.
+   */
+  getCache (key) {
+    return this.rpc.sendAndObserveResponse(
+      'cache',
+      ['get', key]
+    ).pipe(
+      pluck('result')
+    )
+  }
```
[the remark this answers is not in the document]
2color: I had that in mind, but that breaks the API because `state()` is an subscription and `getCache` returns the first value and completes the observable.

packages/aragon-api/src/index.js
```diff
@@ -202,22 +261,43 @@ export class AppProxy {
...
    * Listens for events, passes them through `reducer`, caches the resulting state and re-emits that state for easy chaining.
+   * `cache`ing the `state` key will emit the new state in the application frontend.
    *
-   * This is in fact sugar on top of [`state`](#state), [`events`](#events) and [`cache`](#cache).
+   * For caching purposes the event fetching is split into two steps:
+   *  - Fetching past events with `pastEvents`
+   *  - Subscribing to new events
    *
-   * The reducer takes the signature `(state, event)` à la Redux. Note that it _must always_ return a state, even if it is unaltered by the event.
+   * The reducer takes the signature `(state, event)` and must return a promise that resolves to state, even if it is unaltered by the event.
```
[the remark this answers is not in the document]
2color: You're right. I'll clarify that

packages/aragon-api/src/index.js
```diff
@@ -228,19 +308,74 @@ export class AppProxy {
...
+    // External contracts can specify their own `initializationBlock` which will be used in case the cache is empty
+    const getPastEvents = (cachedFromBlock, toBlock) => merge(
+      this.pastEvents(cachedFromBlock, toBlock),
+      ...externals.map(({ contract, initializationBlock }) => contract.pastEvents(cachedFromBlock || initializationBlock, toBlock))
+    )
+
+    const cachedState$ = this.state().pipe(first())
+    const cachedBlock$ = this.getCache(CACHED_BLOCK_KEY)
+    const latestBlock$ = this.web3Eth('getBlockNumber')
+    const initState$ = init ? from(init()) : from([null])
```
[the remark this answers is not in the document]
2color: Indeed. As you explained in the comment. Addressed in https://github.com/aragon/aragon.js/pull/297/commits/5f9e05387f1bf21bbce6102afdc60cea589fa5ff

packages/aragon-api/src/index.js
```diff
@@ -228,19 +308,74 @@ export class AppProxy {
...
+            this.cache('state', state)
+          })
         )
-      ),
+
+        const currentState$ = pastState$.pipe(
+          // Use the last past state as the initial state for reducing current/future states
+          last(),
+          switchMap(pastState => {
+            return merge(currentEvents$, accounts$).pipe(
```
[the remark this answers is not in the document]
2color: > Can we define currentEvents in this block?

Good idea.

> ⚠️ Also, I think we may have a problem with the accounts$ observable being hot; ideally we would only kick it off here (the old semi-hot behaviour will not be a problem anymore after #305).

We only really care about the last value which `publishReplay` should ensure is available. Do you think it fits better here even if we don't have a hot observable problem? 
I put it in the top block to avoid bloating this part with all the object creation part.

packages/aragon-api/src/index.js
```diff
@@ -228,19 +308,74 @@ export class AppProxy {
...
+        const pastState$ = getPastEvents(cachedBlock, pastEventsToBlock).pipe(
+          // Prevent multiple subscriptions invoking duplicate calls
+          share(),
+          // single emission array of all pastEvents -> flatten to process events
+          flatMap(pastEvents => from(pastEvents)),
+          mergeScan(wrappedReducer, { ...cachedState, ...initState }, 1),
+          // debounce to reduce rendering and caching overthead
+          debounceTime(200),
+          tap((state) => {
+            this.cache(CACHED_BLOCK_KEY, pastEventsToBlock)
```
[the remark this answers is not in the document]
2color: Agreed. I moved in https://github.com/aragon/aragon.js/pull/297/commits/3498e969701cc784bd915c000b479b3d74f6c16d after the `last` call. This fixes the problem as noted in your comment and also ensures we only emit one state object for the pastEvents.

packages/aragon-api/src/index.js
```diff
@@ -228,19 +308,74 @@ export class AppProxy {
...
+      this.events(fromBlock),
+      ...externals.map(({ contract }) => contract.events(fromBlock))
+    )
+
+    // If `cachedFromBlock` is null there's no cache, `pastEvents` will use the initialisationBlock
+    // External contracts can specify their own `initializationBlock` which will be used in case the cache is empty
+    const getPastEvents = (cachedFromBlock, toBlock) => merge(
+      this.pastEvents(cachedFromBlock, toBlock),
+      ...externals.map(({ contract, initializationBlock }) => contract.pastEvents(cachedFromBlock || initializationBlock, toBlock))
+    )
```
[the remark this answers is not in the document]
2color: Yeah good idea!

packages/aragon-api/src/index.js
```diff
@@ -228,19 +308,74 @@ export class AppProxy {
...
+
+    const store$ = combineLatest(cachedState$, initState$, cachedBlock$, latestBlock$).pipe(
+      switchMap(([cachedState, initState, cachedBlock, latestBlock]) => {
+        console.debug('- store - initState', initState)
+        console.debug('- store - cachedState', cachedState)
+        console.debug(`- store - cachedBlock ${cachedBlock} | latestBlock: ${latestBlock}`)
+
+        // The block up to which to fetch past events.
+        // The reduced state up to this point will be cached on every load
+        const pastEventsToBlock = latestBlock - BLOCK_REORG_MARGIN
```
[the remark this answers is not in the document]
2color: Basically we want a positive value with the minimum being 0. Therefore I used `Math.max` so that we actually cache later blocks.

2color:
I just discovered a major bug in the current implementation. Because we rely on the cache key `state` to pass state to the application, we need to introduce a different cache key for the cached state. Otherwise, the block safety margin is not taken into consideration and every reduced state will be cached. 

Right now we conflate the two actions:
- Storing state in cache https://github.com/aragon/aragon.js/blob/2ba723fe3b71781f27056a54c6547217cd785c69/packages/aragon-api/src/index.js#L369
- Passing state to app https://github.com/aragon/aragon.js/blob/2ba723fe3b71781f27056a54c6547217cd785c69/packages/aragon-api/src/index.js#L381

packages/aragon-api/src/index.js
2color: Resolved in https://github.com/aragon/aragon.js/pull/297/commits/5f9e05387f1bf21bbce6102afdc60cea589fa5ff

packages/aragon-api/src/index.js
2color: Not directly related, but perhaps we should a tilde instead of caret for rxjs. I just hit a strange bug (https://jsbin.com/sasumoy/edit?html,js,console). It's already reported: https://github.com/ReactiveX/rxjs/issues/4731

packages/aragon-api/src/index.js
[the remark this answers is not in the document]
sohkai: Perhaps then we could rewrite `getCache()` to use `state()` (pipe in `first()`)?

packages/aragon-api/src/index.js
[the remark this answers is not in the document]
sohkai: Yes, `Math.min()` was the wrong direction :smile:

packages/aragon-api/src/index.js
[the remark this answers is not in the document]
sohkai: Ah yes, that's true! I think with #305 we will be able to just use `app.accounts()` to simplify this whole bit.

sohkai:
> I just discovered a major bug in the current implementation. Because we rely on the cache key state to pass state to the application, we need to introduce a different cache key for the cached state.

I was thinking about this earlier, and ended up writing https://github.com/aragon/aragon.js/issues/294#issuecomment-493501779 in reasoning about this.

I'm curious if you might have noticed something that necessitates keeping the current state and cached state separate? I ultimately ended up thinking that it's not a big deal to keep using the current state cache if developers know that some already-cached events may be repeated (even after these changes, it could still happen if a block reorg occurred when the app is running).

The nice benefit of keeping the current state cache as the "safe cached state", is that state won't appear to be briefly "rolled back" if a user closes the app and re-opens it.

packages/aragon-api/src/index.js
sohkai: Hmm I think this type of bug is going to be pretty hard to avoid via a dependency lock (a patch version fixing a bug could also be buggy); the best we could do is add a `package-lock.json` (which _may_ be possible after upgrading web3.js), and of course, tests.

It seems like you might not have had the latest patch version either (rxjs@6.5.2 seems to have this fixed?), so as well, even with the `package-lock.json` we would've been screwed if it was locked to 6.5.1 🤷‍♂.

2color: Yep. `package-lock.json` is the way forward. You're right that we  would have been screwed if it were locked to 6.5.1 but at least we'd have been able to notice that the upgrade took place.

2color:
> I'm curious if you might have noticed something that necessitates keeping the current state and cached state separate? 

If the state is based on an event that was not included after a re-org, refetching the events from the cached block number doesn't guarantee that the state will not be invalid. 

> The nice benefit of keeping the current state cache as the "safe cached state", is that state won't appear to be briefly "rolled back" if a user closes the app and re-opens it.

With https://github.com/aragon/aragon.js/pull/297/commits/819aeb95be9154cc871f64d74ef177efbc53c6ae the app will still render the latest state. It will just use the cached (slightly older) for refetching and reducing the delta from the last safety margin to the current head.  So we still get most of the benefit without the supposed risk –as I described above–.

2color:
> Secondly, a bit of an API nitpick (and I was originally the one who took the shortcut with external.events() 🙈), in cases where we use getPastEvents() and events(), we should allow the user to specify the full range of options that web3.js supports.

I'll change it to an options object so that it can support `topics` and `filter` too, as per [web3.js](https://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#getpastevents):
```
pastEvents: (options = {})
events: (options = {})
```
That means that calls will look like: pastEvents({ fromBlock: 10, toBlock: 20 }).


That begs the question of what do to do about the API for the app's contract? 
With the introduced change it will feel inconsistent:
```
events (fromBlock)
pastEvents (fromBlock, toBlock)
```

Any thoughts?

2color:
Regarding the `syncStatus` method. Right now the aragonAPI is used by apps and app scripts to communicate with the native context. 
Introducing something like `syncStatus` would be a deviation from that, insofar as it's used to communicate state between the app script and the app.  
Did you have any concrete thoughts on how that would look like?

sohkai:
> To summarise, I'm working under the assumption that anything 100 blocks old could get dropped and not be included in a reorg. If we can safely assume that reducing and caching an event that will get dropped is safe, we don't need 2 cache keys for cached state and current state.

👍 I think this is a good assumption and good trade-off for safety.

packages/aragon-api/src/index.js
```diff
@@ -148,6 +176,24 @@ export class AppProxy {
...
+        const eventArgs = [
+          address,
+          jsonInterface.filter(
+            (item) => item.type === 'event'
+          ),
+          options
+        ]
+
+        return defer(
+          () => this.rpc.sendAndObserveResponse(
```
sohkai: Because we use `from()` in the `pastEvents` handlers, I think we should be using `sendAndObserveResponses()` for these past events instead.

packages/aragon-api/src/index.js
```diff
@@ -181,6 +221,21 @@ export class AppProxy {
     return value
   }
 
+  /**
+   * Get a value from the application cache.
+   *
+   * @param  {string} key   The cache key to get a value for
```
sohkai:
```suggestion
   * @param  {string} key The cache key to get a value for
```

packages/aragon-api/src/index.js
```diff
@@ -144,6 +168,22 @@ export class AppProxy {
...
+      pastEvents: (options = {}) => {
+        const eventArgs = [
+          address,
+          jsonInterface.filter(
+            (item) => item.type === 'event'
+          ),
+          options
+        ]
+
+        return this.rpc.sendAndObserveResponse(
```
sohkai: Here and in `pastEvents()`, shouldn't this be a `sendAndObserveResponses()`, as `@aragon/wrapper` streams individual events (with the `from()`)?

2color: Well that depends on how it behaves when there are two nested `from` calls on a promise. `pastEvents` **only** returns a promise. In all the tests I've done so far this seemed fine. 


https://github.com/aragon/aragon.js/blob/4c22a1173f4d4cfeabba92014b03d1e0f6a25520/packages/aragon-wrapper/src/core/proxy/index.js#L41-L44

packages/aragon-api/src/index.js
```diff
@@ -198,22 +253,50 @@ export class AppProxy {
...
+   * @param  {Object} [options] An optional options object
+   * @param  {Array.<{contract: Object, initializationBlock: String}>} [options.externals] An optional array of objects containing `contract` (as returned from `api.external`) and an optional `initializationBlock` from which to fetch events
+   * @param  {Function} [options.init] An optional initialization function for the state. Should return a promise that resolves to the init state.
+   * @return {Observable} An Observable that emits the application state every time it changes. The type of the emitted values is application specific.
    */
-  store (reducer, events = [empty()]) {
-    const initialState = this.state().pipe(first())
+  store (reducer, { externals = [], init } = {}) {
+    const CACHED_BLOCK_KEY = 'CACHED_BLOCK_KEY'
+    const CACHED_STATE_KEY = 'CACHED_STATE_KEY'
```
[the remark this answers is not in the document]
2color: Now that we've agreed on having a separate cached state key it makes sense. They were kept separate as the initial implementation maintained inter-operability with the standard `state` key

packages/aragon-api/src/index.js
sohkai: Ahh gotcha, I see what's going on. I originally thought it was streaming the events one by one, but I think this makes more sense since it more closely aligns with the web3.js API anyway (all at once) 👍.

PR #301, opened 2019-05-14
Log errors thrown in reducer

### LINKED ISSUE #262: Errors inside app.store reducers are not thrown
Errors occurried inside store reducers are not shown in console. The following code fails silently, for example:

```js
app.store(
  (state, event) => {
    throw new Error('hi')
    return state
  }
)
```

On the other hand, if we put an Error in the reducer used in the [test](https://github.com/aragon/aragon.js/blob/master/packages/aragon-api/src/index.test.js#L235), it is thrown as expected. I don't know why.

Anyone have any idea why errors are dismissed inside app.store?

2color:
Fixes #262

### DIFF
--- a/packages/aragon-api/src/index.js
+++ b/packages/aragon-api/src/index.js
@@ -228,7 +228,9 @@ export class AppProxy {
     // (because of the `Promise.resolve`).
     const wrappedReducer = (state, event) =>
       from(
-        Promise.resolve(reducer(state, event))
+        Promise.resolve(reducer(state, event)).catch(e => console.error(
+          `Error from app reducer:`, e)
+        )
       )
 
     const store$ = initialState.pipe(
--- a/packages/aragon-api/src/index.test.js
+++ b/packages/aragon-api/src/index.test.js
@@ -206,7 +206,7 @@ test('should return the state from cache', t => {
   })
 })
 
-test('should create a store reducer', async t => {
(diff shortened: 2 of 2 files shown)


packages/aragon-api/src/index.js
```diff
@@ -228,7 +228,9 @@ export class AppProxy {
     // (because of the `Promise.resolve`).
     const wrappedReducer = (state, event) =>
       from(
-        Promise.resolve(reducer(state, event))
+        Promise.resolve(reducer(state, event)).catch(e => console.error(
+          `Error from app reducer:`, e)
```
sohkai: I'm wondering what the correct behaviour for us should be. Should we re-throw the error, continue like nothing happened (as we do here, but return `undefined` for the new state), or try to give the error back to the frontend.

I'd recommend at least rethrowing the error for now, so the observable stream stops where it is, but we should also think about having a mechanism for frontends to detect their reducer has failed.

2color: Good point. I didn't think that far as I was a little averse to API changes and wanted to get logging before improving the mechanism. Nevertheless, this is worth looking deeper into. 

I'll do some more testing and report back.

2color:
Closing in favour of https://github.com/aragon/aragon.js/pull/313

PR #303, opened 2019-05-16
Wrapper: fix initAcl() on localhost

0x6431346e:
**Changes**

- Moved the `getBlockNumber()` call higher

**Why**

This seems to fix the issue with `rc.6` where the wrapper would initialize, but never emit permissions/apps/etc.

**How**

This might be related to ganache, web3 or `Proxy.events`. It looks like when another web3 function is used between calling `Proxy.events` and connecting/subscribing, it will no longer emit:
```
   const someEvents = this.aclProxy.events()
   console.log('acc', await this.web3.eth.getAccounts())
   someEvents.subscribe(console.log)
```
(it emits if I remove the `getAccounts()` call)

Note: on rinkeby it works without this change 🤷‍♂ 

- [ ] I have updated the associated documentation with my changes

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -241,6 +241,9 @@ export default class Aragon {
 
     const REORG_SAFETY_BLOCK_AGE = 100
 
+    const currentBlock = await this.web3.eth.getBlockNumber()
+    const cacheBlockHeight = currentBlock - REORG_SAFETY_BLOCK_AGE
+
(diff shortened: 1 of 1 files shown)


sohkai:
🤦‍♂ This is crazy

sohkai:
@0x6431346e This has now been released as `@aragon/wrapper@5.0.0-rc.7`

0x6431346e:
> @0x6431346e Was this happening consistently on every load?

Yes

PR #304, opened 2019-05-17
Clean up API documentation

sohkai:
Adds a bit of consistency, updates links, and reorders sections to place more useful APIs closer to the top. Also uses `api` as the recommended name, to align with `api-react`.

Also fixes https://github.com/aragon/aragon.js/issues/302.

PR #305, opened 2019-05-18
messenger: use defer() for observed requests to avoid dropping responses

sohkai:
As their name implies, `sendAndObserveResponses()` and `sendAndObserveResponse()` only work when they have a subscriber attached to receive the response, otherwise we result in the current confusing "semi-hot" behaviour where an RPC request is initiated immediately (from the app) but whose response (from the client) gets dropped because there's no observer set on the app.

This changes the behaviour of these RPC requests to always make them cold, such that requests are only kicked off once the first subscriber is attached.

**Todo**:

- [x] Add a note to the API docs about this behaviour after https://github.com/aragon/aragon.js/pull/304 is merged.
- [x] Fix tests

### DIFF
--- a/docs/APP.md
+++ b/docs/APP.md
@@ -92,12 +92,17 @@ const frontendOfApp = new AragonApp(new providers.WindowMessage(window.parent))
 ```
 
 > **Note**<br>
-> Most of the returned observables will propagate errors from `@aragon/wrapper` (e.g. the Aragon client) if an RPC request failed. An example would be trying to use `api.call('nonexistentFunction')`.
+> Most of the returned observables will propagate errors from `@aragon/wrapper` (e.g. the Aragon client) if an RPC request failed. An example would be trying to use `api.call('nonexistentFunction')`. Multi-emission observables (e.g. `api.accounts()`) will forward the error without stopping the stream, leaving the subscriber to handle the error case.
 
 > **Note**<br>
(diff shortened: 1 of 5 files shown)


2color:
I think this is a good idea. For the sake of clarity, it's worth noting that the current behaviour is simply *hot*.  The problem with events getting dropped is because it's hot, isn't it?

Wit this change it's up to the callers to make sure they use `publish` or `share` if they have multiple subscriptions, otherwise each will instantiate a new observable and a corresponding RPC request.

sohkai:
> For the sake of clarity, it's worth noting that the current behaviour is simply hot. The problem with events getting dropped is because it's hot, isn't it?

Kind of... the problem with the messenger is that we're going across an RPC wire. Everything is _hot_ on the way across the wire, but on the way back, we need to subscribe otherwise we will never be [running the filter for events](https://github.com/aragon/aragon.js/blob/master/packages/aragon-rpc-messenger/src/index.js#L94).

[`provider.messages$`](https://github.com/aragon/aragon.js/blob/master/packages/aragon-rpc-messenger/src/providers/WindowMessage.js#L20) is hot, but the rest of the operator chain for a particular request's response won't be run unless it's been subscribed to.

> Wit this change it's up to the callers to make sure they use publish or share if they have multiple subscriptions,

Yes, although I think this is the current behaviour now anyway (a new observable is always instantiated and returned upon an RPC request). For some long-lived multi-emission APIs, like `accounts()`, we could set it up with a `publishReplay()` to dedupe requests (although I'm not super worried about this right now given how cheap these observables should be).

2color:
Righto. That's all clear now.

PR #306, opened 2019-05-18
wrapper: clamp ACL cached block height to 0 for safety in local chains

sohkai:
On local chains (e.g. aragen's), there may not have been 100 blocks yet 😅.

cc @0xGabi, not sure if this fixes the local chain issue you were seeing but will keep investigating.

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -242,7 +242,7 @@ export default class Aragon {
     const REORG_SAFETY_BLOCK_AGE = 100
 
     const currentBlock = await this.web3.eth.getBlockNumber()
-    const cacheBlockHeight = Math.min(currentBlock - REORG_SAFETY_BLOCK_AGE, 0) // clamp to 0 for safety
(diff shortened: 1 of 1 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -242,7 +242,7 @@ export default class Aragon {
     const REORG_SAFETY_BLOCK_AGE = 100
 
     const currentBlock = await this.web3.eth.getBlockNumber()
-    const cacheBlockHeight = currentBlock - REORG_SAFETY_BLOCK_AGE
+    const cacheBlockHeight = Math.min(currentBlock - REORG_SAFETY_BLOCK_AGE, 0) // clamp to 0 for safety
```
```suggestion
    const cacheBlockHeight = Math.max(currentBlock - REORG_SAFETY_BLOCK_AGE, 0) // clamp to 0 for safety
```

We probably want to use `max` as min will always take `0` when `0` and `7801359` are passed in.

sohkai: Good catch 😄

PR #308, opened 2019-05-22
Add handler for searching local identities 

2color:
## TODO
- [x] Agree on API 
- [x] Add search method to local identity provider

## Related issue/PRs
- https://github.com/aragonone/product/issues/59
- https://github.com/aragon/aragon-apps/pull/862


packages/aragon-wrapper/src/rpc/handlers/search-identities.js
```diff
@@ -0,0 +1,11 @@
+export default function (request, proxy, wrapper) {
+  const [searchTerm] = request.params
+
+  if (searchTerm.length < 3) {
+    return Promise.reject(
```
AquiGorka: Why `reject` over answering with an empty array?

2color: After thinking, for no good reason.

packages/aragon-wrapper/src/rpc/handlers/search-identities.js
```diff
@@ -0,0 +1,10 @@
+export default function (request, proxy, wrapper) {
+  const [searchTerm] = request.params
+
+  if (searchTerm.length < 3) {
```
AquiGorka: Do we need this here too? We do it in the frontend and in the `IdentityProvider` too

2color: Good question. I think that this is an important place to keep it because it's the gateway to all providers. Perhaps it's in the providers that it isn't necessary to implement this.

AquiGorka: https://github.com/aragon/aragon.js/pull/322

PR #310, opened 2019-05-27
Modify all events APIs to allow passing options objects

sohkai:
Changes all of the `.events()` and `.pastEvents()` APIs to accept an options object, rather than explicit from/to blocks.

This change also affects the API for `Proxy.events()`: it now defaults `fromBlock` to the app's initialization block unless the value is explicitly overridden by the user, staying consistent with the rest of the APIs (e.g. how the [external handlers handle the options](https://github.com/aragon/aragon.js/compare/app-state-caching...api-events-options?expand=1#diff-e343de8b16c983166963b5cb47f0e383R37)). The only place where this should be a breaking change is in the [repo events subscription](https://github.com/aragon/aragon.js/compare/app-state-caching...api-events-options?expand=1#diff-5b914b5b526a57f89974fa2fe009ee80R439), where I originally added the `fromBlock` functionality.

### DIFF
--- a/packages/aragon-api/src/index.js
+++ b/packages/aragon-api/src/index.js
@@ -134,7 +134,7 @@ export class AppProxy {
    * @return {Observable} An [RxJS observable](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) that emits [Web3 events](https://web3js.readthedocs.io/en/1.0/glossary.html#specification).
    */
   pastEvents (options) {
-    return this.rpc.sendAndObserveResponses(
+    return this.rpc.sendAndObserveResponse(
       'past_events',
       [options]
     ).pipe(
--- a/packages/aragon-api/src/index.js
+++ b/packages/aragon-api/src/index.js
@@ -113,13 +113,14 @@ export class AppProxy {
   /**
    * Listens for events on your app's smart contract from the last unhandled block.
    *
-   * @param  {string} fromBlock block from which to fetch the events
+   * @param  {object} [options] web3.eth.Contract.events()' options
+   *   If not given, defaults fromBlock to this app's initializationBlock
    * @return {Observable} An [RxJS observable](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) that emits [Web3 events](https://web3js.readthedocs.io/en/1.0/glossary.html#specification).
    */
-  events (fromBlock) {
+  events (options) {
     return this.rpc.sendAndObserveResponses(
       'events',
-      [fromBlock]
+      [options]
     ).pipe(
       pluck('result')
     )
@@ -128,14 +129,14 @@ export class AppProxy {
   /**
    * Fetch past events from your app's smart contract for requestsed range
    *
-   * @param  {string} fromBlock block from which to fetch the events
-   * @param  {string} toBlock block up to which to fetch the events
+   * @param  {object} [options] web3.eth.Contract.events()' options
+   *   If not given, defaults fromBlock to this app's initializationBlock
    * @return {Observable} An [RxJS observable](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) that emits [Web3 events](https://web3js.readthedocs.io/en/1.0/glossary.html#specification).
    */
-  pastEvents (fromBlock, toBlock) {
-    return this.rpc.sendAndObserveResponse(
+  pastEvents (options) {
+    return this.rpc.sendAndObserveResponses(
       'past_events',
-      [fromBlock, toBlock]
+      [options]
     ).pipe(
       pluck('result')
     )
@@ -147,29 +148,14 @@ export class AppProxy {
    *
    * @param  {string} address The address of the external contract
    * @param  {Array<Object>} jsonInterface The [JSON interface](https://web3js.readthedocs.io/en/1.0/glossary.html#glossary-json-interface) of the external contract.
-   * @return {Object}  An external smart contract handle. Calling any function on this object will send a call to the smart contract and return an [RxJS observable](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) that emits the value of the call.
+   * @return {Object}  An external smart contract handle, containing the following methods:
+   *   - `events(options)`: returns a multi-emission observable with individual events found
+   *   - `pastEvents(options)`: returns a multi-emission observable with individual events found in past blocks
+   *   - Calling any other method on the handle will send a call to the smart contract and return an single emission observable with the result.
    */
   external (address, jsonInterface) {
     const contract = {
-      events: (fromBlock) => {
-        const eventArgs = [
-          address,
-          jsonInterface.filter(
-            (item) => item.type === 'event'
-          )
-        ]
-        if (typeof fromBlock === 'number') {
-          eventArgs.push(fromBlock)
-        }
-
-        return this.rpc.sendAndObserveResponses(
-          'external_events',
-          eventArgs
-        ).pipe(
-          pluck('result')
-        )
-      },
-      pastEvents: (options = {}) => {
+      events: (options) => {
         const eventArgs = [
           address,
           jsonInterface.filter(
@@ -178,7 +164,23 @@ export class AppProxy {
           options
         ]
 
-        return this.rpc.sendAndObserveResponse(
+        return this.rpc.sendAndObserveResponses(
+          'external_events',
+          eventArgs
+        ).pipe(
+          pluck('result')
+        )
+      },
+      pastEvents: (options) => {
+        const eventArgs = [
+          address,
+          jsonInterface.filter(
+            (item) => item.type === 'event'
+          ),
+          options
+        ]
+
+        return this.rpc.sendAndObserveResponses(
           'external_past_events',
           eventArgs
         ).pipe(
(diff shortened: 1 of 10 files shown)


packages/aragon-api/src/index.js
```diff
@@ -128,14 +129,14 @@ export class AppProxy {
...
-   * @param  {string} fromBlock block from which to fetch the events
-   * @param  {string} toBlock block up to which to fetch the events
+   * @param  {object} [options] web3.eth.Contract.events()' options
+   *   If not given, defaults fromBlock to this app's initializationBlock
    * @return {Observable} An [RxJS observable](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) that emits [Web3 events](https://web3js.readthedocs.io/en/1.0/glossary.html#specification).
    */
-  pastEvents (fromBlock, toBlock) {
-    return this.rpc.sendAndObserveResponse(
+  pastEvents (options) {
+    return this.rpc.sendAndObserveResponses(
```
2color:
```suggestion
    return this.rpc.sendAndObserveResponse(
```

packages/aragon-api/src/index.js
```diff
@@ -178,7 +180,7 @@ export class AppProxy {
           options
         ]
 
-        return this.rpc.sendAndObserveResponse(
+        return this.rpc.sendAndObserveResponses(
```
2color:
```suggestion
        return this.rpc.sendAndObserveResponse(
```

packages/aragon-api/src/index.test.js
```diff
@@ -133,8 +133,31 @@ test('should return the events observable', t => {
...
+  // arrange
+  const pastEventsFn = Index.AppProxy.prototype.pastEvents
+  const observable = of({
+    id: 'uuid1',
+    result: ['eventA', 'eventB']
+  })
+  const instanceStub = {
+    rpc: {
+      // Mimic behaviour of @aragon/rpc-messenger
+      sendAndObserveResponses: createDeferredStub(observable)
```
2color:
```suggestion
      sendAndObserveResponse: createDeferredStub(observable)
```

packages/aragon-api/src/index.test.js
```diff
@@ -133,8 +133,31 @@ test('should return the events observable', t => {
...
+    }
+  }
+  // act
+  const result = pastEventsFn.call(instanceStub)
+  // assert
+  result.subscribe(value => {
+    t.deepEqual(value, ['eventA', 'eventB'])
+  })
+  t.is(instanceStub.rpc.sendAndObserveResponses.getCall(0).args[0], 'past_events')
```
2color:
```suggestion
  t.is(instanceStub.rpc.sendAndObserveResponse.getCall(0).args[0], 'past_events')
```

packages/aragon-api/src/index.test.js
```diff
@@ -159,14 +182,26 @@ test('should return an handle for an external contract events', t => {
...
+      eventsCall.args[1],
+      ['0xextContract', [jsonInterfaceStub[0]], { fromBlock: 2000 }]
+    )
+  })
+  // past events from block 3000 (just re-use the same events as `events` for simplicity)
+  result.pastEvents({ fromBlock: 3000 }).subscribe(value => {
     t.deepEqual(value, { name: 'eventA', value: 3000 })
 
-    t.is(instanceStub.rpc.sendAndObserveResponses.getCall(0).args[0], 'external_events')
+    const pastEventsCall = instanceStub.rpc.sendAndObserveResponses.getCall(1)
```
2color:
```suggestion
    const pastEventsCall = instanceStub.rpc.sendAndObserveResponse.getCall(1)
```

packages/aragon-wrapper/src/core/proxy/index.js
```diff
@@ -1,6 +1,21 @@
 import { fromEvent, from } from 'rxjs'
 import { filter } from 'rxjs/operators'
 
+function getEventNames (eventNames) {
+  // Get all events
+  if (!eventNames) {
+    eventNames = ['allEvents']
```
2color: What do you think about just returning rather than assigning a new value:
```suggestion
    return ['allEvents']
```

packages/aragon-wrapper/src/core/proxy/index.js
```diff
@@ -16,46 +31,39 @@ export default class Proxy {
...
-    if (!Array.isArray(eventNames)) {
-      eventNames = [eventNames]
-    }
+  pastEvents (eventNames, options = {}) {
+    eventNames = getEventNames(eventNames)
+    options.fromBlock = options.fromBlock || this.initializationBlock
 
     if (eventNames.length === 1) {
-      //
+      // Get a specific event
```
2color:
```suggestion
      // Get a specific or all events
```

2color: I'd also add this comment below to `event()`

packages/aragon-wrapper/src/index.js
```diff
@@ -435,8 +435,8 @@ export default class Aragon {
     // These may modify the implementation addresses of the proxies (modifying their behaviour), so
     // we invalidate any caching we've done
     const updatedApps$ = this.kernelProxy
-      // Override events subscription with empty options to subscribe from latest block
-      .events('SetApp', {})
+      // Only need to subscribe from latest block
+      .events('SetApp', { fromBlock: 'latest' })
```
2color: From [event](https://web3js.readthedocs.io/en/1.0/web3-eth-contract.html#contract-events) the docs say that `fromBlock` only accepts a only a `number`. In other web3 methods it accepts a `string`. 
Have you tested this?

packages/aragon-wrapper/src/rpc/handlers/events.js
```diff
@@ -1,8 +1,7 @@
 export default function (request, proxy) {
-  const fromBlock = request.params && request.params[0]
-
-  if (fromBlock) {
-    return proxy.events(null, { fromBlock })
+  const [eventsOptions] = request.params
+  if (eventsOptions != null) {
```
2color: Was the loose comparison deliberate to get undefined too?

2color:
> This change also affects the API for Proxy.events(): it now defaults fromBlock to the app's initialization block unless the value is explicitly overridden by the us

That was already the case. We were defaulting to the contract's init block with default params: https://github.com/aragon/aragon.js/pull/310/files#diff-13b1c645b72103672efd438469496c7cL48

packages/aragon-wrapper/src/index.js
sohkai: Not explicitly, but we should! I remember the web3.js code itself checking for 'latest'.

packages/aragon-wrapper/src/rpc/handlers/events.js
sohkai: Yes, in case that last param wasn't given it would be undefined.

sohkai:
> For for pastEvents I'd change the observables to single emissions ones, sendAndObserveResponse

Ah yes, forgot to revert this one after learning about what the API of `pastEvents()` actually was :smile:.

> We were defaulting to the contract's init block with default params: 

Yes, this is a bit subtle; previously it would **only** default if you didn't pass along an options object. Now it'll default the `fromBlock` option as long as you don't explicitly provide an options object with the `fromBlock` option set.

packages/aragon-wrapper/src/rpc/handlers/events.js
```diff
@@ -1,8 +1,7 @@
 export default function (request, proxy) {
-  const fromBlock = request.params && request.params[0]
```
sohkai: Ahh shoot, we need to be backwards compatible here with `@aragon/api@1`, as old apps will still be using the first parameter as the `fromBlock`.

2color: I thought this PR was part of a major release, which'll break API

sohkai: Unfortunately `@aragon/wrapper` will have to stay compatible with older versions of `@aragon/api`, since it acts as the "server" :(

2color: So essentially all DAOs, including ones with non upgraded apps using the old API will be affected by an upgrade to `@aragon/wrapper` because it's imported by `aragon/aragon` and `apm-serve`: 

https://github.com/aragon/apm-serve/blob/b7dfacd084f60b1fa9e1cf01aa459004d3f06d54/lib/apm-router.js#L39-L41

will always load the latest?

sohkai: Yes.

packages/aragon-wrapper/src/index.js
sohkai: It turns out that in web3.js 2.x, [`'latest'` is checked explicitly](https://github.com/ethereum/web3.js/blob/2.x/packages/web3-core-subscriptions/src/subscriptions/eth/LogSubscription.js#L53) but in web3.js 1.x, ['latest' implicitly skips fetching the past logs since it's not a finite number](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-core-subscriptions/src/subscription.js).

The [EIP specifying the RPC calls does note 'latest' as a possible value, amongst others](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1474.md#eth_getlogs), and the [wiki's json-rpc entry on `getLogs`](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getlogs) notes that 'latest' should be the default value for `fromBlock` (and indeed, web3.js is programmed with that assumption).

Even though web3.js 1.x doesn't handle `'latest'` in the _best_ way, given that it is still _working_ and `'latest'` is a possible value for the JSON-RPC API's `fromBlock`, I'm inclined to keep this as it's explicit and any potential regression would squarely be a bug in web3.js.

sohkai:
Superceded by #367.

Issue #229, opened 2018-12-26
RPC: Potential memory leak from multi-emission subscriptions


kernelwhisperer:
There is no way to unsubscribe after doing `sendAndObserveResponses`.

For example, sending:
```
{
      id: 1,
      jsonrpc: '2.0',
      method: 'accounts'
}
```
will trigger the wrapper to *always* send updates with `id: 1`.

I suppose this could be be a common pattern:
```
onButtonClick () {
  const accounts = await app.accounts.toPromise()
  ...
}
```
which will make the wrapper send updates for each JSONRPC request, even though the client would expect a one time response.

I can't really find the pub/sub spec :thinking: , but it should be something like this: https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB

sohkai:
One thought on this, we could wrap `sendAndObserveResponses()` with an `endWith`, sending a message to the wrapper to also end the stream on its side.

PR #311, opened 2019-05-27
messenger: complete multi-emission response streams

sohkai:
Previously, if an RPC endpoint returned a multi-emission response (e.g. `api.pastEvents()`), the user would never know when it was completed, despite `@aragon/wrapper` completing the stream on its side.

It wasn't an issue before, since all of our multi-emission responses were meant to be infinite, but with `pastEvents()` this is now a problem.

This change to `@aragon/wrapper` should be backwards compatible to those who are still using `@aragon/api@v1`, since that version of the API only includes infinite multi-emission responses and all others are single-emission responses that already terminate due to a `first()`.

This also mitigates the receiving side of #229, but does not solve it completely (we would need to send an 'unsubscribe' event back to the RPC handler to mitigate it completely).


sohkai:
Deferring this for later; it's unclear exactly if this is necessary and what impact it would have.

sohkai:
It turns out `api.pastEvents()` is a single-emission result, not a multi-emission one, so the root reason behind this change is moot.

However, that being said, I think it's still useful functionality to have in the messenger layer, to cover our backs when we need a stream that does end.

packages/aragon-rpc-messenger/src/signals.js
```diff
@@ -0,0 +1,5 @@
+export const COMPLETE = Symbol('RPC_COMPLETE')
```
2color: Could this pose problems given that symbols can't be serialised for `postMessage`?
https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm

sohkai: Ah yes, this was designed around that; the receiver actually imports this symbol from `@aragon/rpc-messenger` and uses it directly within their context, so it never gets sent through a `postMessage`.

sohkai:
Superceded by #367.

Issue #119, opened 2018-06-05
Add ability for frontend to trigger events in the script


sohkai:
It'd be great if there was a way for the frontend to send arbitrary events to the script, and define logic in the script to handle it.

It could look like `app.trigger('<event name>', <event params...>)` from the frontend, and `@aragon/client`'s `app.store()` could receive these as-is.

See https://github.com/aragon/aragon.js/pull/313#issuecomment-496577875 for an description of how this could potentially be done.

2color:
That's an interesting idea. Can you give an example of a use-case?

sohkai:
A few come to mind:

- Fetching more data (if, in the future, reducing all of the state takes too long)
- Triggering an external API call that is better handled as part of the webworker
- Optimistic UI for actions (multiple ways to do this, but "triggering" an "optimistic" event in the script could be one way)

PR #313, opened 2019-05-27
App state caching: improvements

sohkai:
A couple of improvements to #297.

- 956d722: Now that #305 is merged, we can use `this.accounts()` directly without worrying about the first emission being lost
- d307e44: Supersedes #301; the `new Promise()` also catches the case where `reducer()` throws synchronously, and the `catchError()` operator allows us to log and re-throw the error
  - To be on the safe side, opted to re-throw the error and end the `store()` stream for now, rather than silently returning the previous `state` and continuing the reducing
- b451ece: Allow any state reduced from past events to be propagated as the current state, now that the cached state is kept in a separate key
  - In the future, we may want to completely move the `state` to be in-memory, since now we're effectively using double the storage (`state` and `cache` representations of the app's reduced state)

### DIFF
--- a/packages/aragon-api/src/index.js
+++ b/packages/aragon-api/src/index.js
@@ -1,7 +1,6 @@
 import { forkJoin, from, merge } from 'rxjs'
 import {
   catchError,
-  debounceTime,
   endWith,
   flatMap,
   filter,
@@ -10,6 +9,7 @@ import {
   last,
   pluck,
   publishReplay,
+  sampleTime,
   startWith,
   switchMap,
   tap
@@ -363,8 +363,9 @@ export class AppProxy {
         console.debug(`- store - currentEvents$: from: ${pastEventsToBlock} -> future`)
 
         return getPastEvents(cachedBlock, pastEventsToBlock).pipe(
-          // throttle to reduce rendering and caching overthead
           mergeScan(wrappedReducer, { ...cachedState, ...initState }, 1),
+          // throttle to reduce rendering and caching overthead
+          sampleTime(200),
           tap((state) => {
             this.cache('state', state)
             console.debug('- store - reduced state from past event:', state)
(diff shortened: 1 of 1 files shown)


packages/aragon-api/src/index.js
```diff
@@ -284,33 +298,22 @@ export class AppProxy {
...
     const wrappedReducer = (state, event) =>
       from(
-        // Ensure a promise is returned even if the reducer returns an array
-        Promise.resolve(reducer(state, event)).catch(e => console.error(
-          `Error from app reducer:`, e)
-        )
+        // Ensure a promise is returned even if the reducer returns an array or throws
+        new Promise((resolve) => resolve(reducer(state, event)))
+      ).pipe(
+        catchError((err) => {
```
2color: Nice. I wasn't aware of this operator

packages/aragon-api/src/index.js
```diff
@@ -360,6 +363,10 @@ export class AppProxy {
 
         return getPastEvents(cachedBlock, pastEventsToBlock).pipe(
           mergeScan(wrappedReducer, { ...cachedState, ...initState }, 1),
+          tap((state) => {
```
2color: I would put this after the `last` operator. The pastEvents come all at once, therefor I'd first reduce them all and then render. This could make it a little sluggish.

2color:
> 956d722: Now that #305 is merged, we can use this.accounts() directly without worrying about the first emission being lost

Didn't see this and already implemented it in #297 

> To be on the safe side, opted to re-throw the error and end the store() stream for now, rather than silently returning the previous state and continuing the reducing

Sensible

> In the future, we may want to completely move the state to be in-memory, since now we're effectively using double the storage (state and cache representations of the app's reduced state)

I agree. How do you envision the passing of state between script to app?

sohkai:
> I agree. How do you envision the passing of state between script to app?

Good question! Been thinking a little bit about how this would work, since it'd also be require for the ["triggers"](#119).

This state "trigger" could be implemented directly as a `ReplaySubject(1)`, held in a mapping based on the app's proxy address. Any time you send a response to it, it would do a `.next()` on the Subject. If you request it, you get subscribed to the relevant subject (again, based on the sender's proxy address).

The ephemeral "trigger" would be exactly the same as above, but just using a normal `Subject()`.

packages/aragon-api/src/index.js
sohkai: Opted to throttle these updates with a `sampleTime()`

PR #314, opened 2019-05-27
Feat/Forwarded Actions API

topocount:
**Changes**

This PR implements an interface that:
1. Allows Forwarders to log forwarded actions that were caught and update them as they are either approved or declined: see `newForwardedAction` and `updateForwardedAction`
2. Allows target apps to track the progress of forwarded actions logged by Forwarders and handle them accordingly in script.js and other webWorker files for the purpose of informing the user of relevant activity elsewhere: see `getForwardedActions`

`newForwardedAction` and `updateForwardedAction` are really syntactic sugar on top of the `setForwardedAction` wrapper method. 

`getForwardedActions` is intended to be implemented similarly to an external contract event subscription in script.js. The key difference though is that, rather than just emitting a single action on each change. All forwarded actions targetting the subscribed app and their present status are emitted to the app. The idea behind this is that all state can be naively updated after one change, as opposed to the app needing to completely rebuild state using events each time it loads. Because forwarded actions are essentially managed by the wrapper, it makes sense that the wrapper also serves as a single source of truth for forwarded action state, as opposed to invidual apps each compiling it's own state. 

- [X] I have updated the associated documentation with my changes


packages/aragon-api/src/index.js
```diff
@@ -538,6 +556,76 @@ export class AppProxy {
...
+   */
+  newForwardedAction (actionId, evmScript) {
+    return this.rpc.send(
+      'update_forwarded_action',
+      [actionId, evmScript]
+    )
+  }
+
+  /**
+   * update a forwarded action's state the wrapper
```
sohkai:
```suggestion
   * Update the state of a forwarded action
```

packages/aragon-api/src/index.js
```diff
@@ -538,6 +556,76 @@ export class AppProxy {
...
+   *
+   * If the execution is pending, state = 0
+   * If the execution completed, state = 1
+   * If the execution failed, state = 2
+   *
+   * @param {string} actionId ID assigned to forwarded action in Forwarder
+   * @param {integer} state The current state of the forwarded action within the forwarder
```
sohkai: I'm not super happy with this interface; we should probably use string constants instead (can't use symbols because they don't go over the RPC wire).

packages/aragon-api/src/index.js
```diff
@@ -538,6 +556,76 @@ export class AppProxy {
...
+   * @param {string} [evmScript=''] The optionally updated execution script
```
sohkai: The EVMScript being updated here would be a case like DotVoting manipulating the execution script based on the voting results?

Also, maybe a better default is `null` (`undefined` also doesn't go through the RPC wire).

packages/aragon-api/src/index.js
```diff
@@ -538,6 +556,76 @@ export class AppProxy {
...
+   */
+  updateForwardedAction (actionId, state, evmScript = '') {
+    return this.rpc.send(
+      'update_forwarded_action',
+      [actionId, evmScript, state]
+    )
+  }
+
+  /**
+   * Listens for forwarded actions that target your app and emits them like contract events
```
sohkai: Perhaps better terminology would be "subscribe"?

packages/aragon-api/src/index.js
```diff
@@ -538,6 +556,76 @@ export class AppProxy {
...
+   *
+   * @return {Observable} An [RxJS observable](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html) that emits an array of forwarded action objects
+   */
+  getForwardedActions () {
+    return this.rpc.sendAndObserveResponses(
+      'get_forwarded_actions'
+    ).pipe(pluck('result'))
```
sohkai: Bit of a nitpick, but we usually do write this as

```
.pipe(
  pluck('result')
)
```

Soon prettier will allow us to avoid this type of bikeshedding, but for now let's keep it consistent :).

packages/aragon-api/src/index.js
```diff
@@ -538,6 +556,76 @@ export class AppProxy {
...
+  }
+
+  /**
+   * Register data for external apps
+   *
+   * @param {string} from Address of the application generating the data
```
sohkai: `from` doesn't seem to be in the function parameters

packages/aragon-api/src/index.js
```diff
@@ -538,6 +556,76 @@ export class AppProxy {
...
+   * @param {<Array>string} to Optional list of addresses of the applications allowed access to the data, defaults to '*'
```
sohkai:
```suggestion
   * @param {Array<string>} [to=['*']] List of applications allowed access to the data, based on their address
```

Also, should `to` be the second parameter (instead of last)? It'd be good to group it with `from` if `from` is supposed to be in here.

packages/aragon-api/src/index.js
```diff
@@ -538,6 +556,76 @@ export class AppProxy {
...
+   * @param {string} dataId internal ID assigned to the data by the originator
+   * @param {string} cid external identifier (e.g., IPFS hash)
```
sohkai: I don't think using CID is very clear; perhaps "contentId" would be better?

packages/aragon-wrapper/src/rpc/handlers/update-forwarded-action.js
```diff
@@ -0,0 +1,15 @@
...
+  ] = request.params
+
+  wrapper.setForwardedAction(
+    proxy.address,
+    actionId,
+    evmScript,
+    state
+  )
+  return Promise.resolve()
+}
```
sohkai: Can we combine this with `get-forwarded-actions.js` into a `forward-actions.js` with multiple exports (e.g. similar to `external.js`)?

packages/aragon-wrapper/src/rpc/handlers/get-forwarded-actions.js
```diff
@@ -0,0 +1,22 @@
...
+    actionsArray
+      .filter(action => (action.target === proxy.address))
+  )
+
+  return wrapper.forwardedActions.pipe(
+    // transform the observable into an event-like object
+    // that only contains actions targeting the proxy
+    map(actions => ({
+      event: 'ForwardedActions',
+      returnValues: getProxyActions(actions)
```
sohkai: I wouldn't do this transformation here. If you need it inside the script (not sure if I would recommend it; perhaps this could be a frontend only thing? I still need to think a bit on this), we can add it to `@aragon/api` like we do with accounts.

packages/aragon-wrapper/src/index.js
```diff
@@ -828,6 +831,130 @@ export default class Aragon {
...
+            // only update any state if the state update is the latest
+            if (actions[updateIndex].state < state) {
+              actions[updateIndex].state = state
+              // only update the evmScript if it's included
+              if (evmScript !== '') actions[updateIndex].evmScript = evmScript
+            }
+            return actions
+          }
+        },
+        [] // actions seed
```
sohkai: Rather than using an array here... can we just use a key value mapping (e.g. `Map` or an object)?

Especially since the `BehaviourSubject` is initialized with an object.

packages/aragon-wrapper/src/index.js
```diff
@@ -828,6 +831,130 @@ export default class Aragon {
...
+  initForwardedActions () {
+    this.forwardedActions = new BehaviorSubject({}).pipe(
+      scan(
+        (actions, { currentApp, actionId, evmScript = '', state = 0 }) => {
+          const updateIndex = actions.findIndex(
+            action => action.currentApp === currentApp && action.actionId === actionId
+          )
+          if (updateIndex === -1) {
+            // set the last address as the target of the forwarded action
+            const target = evmScript ? this.decodeTransactionPath(evmScript).pop().to : ''
```
sohkai: Not 100% sure this works as intended yet; I think `decodeTransactionPath()` only does one layer of EVMScript "unwrapping" (we will soon be adding in multiple layers).

Right now, if you have a flow like `Token Manager -> Voting -> Finance`, I think this would only get you the Voting app.

packages/aragon-wrapper/src/index.js
```diff
@@ -828,6 +831,130 @@ export default class Aragon {
...
+            currentApp && actions.push({ currentApp, actionId, target, evmScript, state })
```
sohkai: Let's use an `if` here instead

packages/aragon-wrapper/src/index.js
```diff
@@ -828,6 +831,130 @@ export default class Aragon {
...
+      evmScript,
+      state
+    })
+  }
+
+  /**
+   * Initialize the appMetadata observable
+   */
+  initAppMetadata () {
+    this.appMetadata = new BehaviorSubject({}).pipe(
```
sohkai: Should this also do a `scan`, to keep all the metadata published in memory?

topocount:
@sohkai,as discussed, our agreed path forward on each discussion point is below:
>We need to take into account what happens when an app's state is cached (in v2 of `@aragon/api`). This means on the next load, only new events will get loaded. We'll likely need to use a caching mechanism for the forwarded actions too, such that we load the initial `scan` state from cache and keep writing to it on each new forwarded action.

We'll implement caching in the wrapper for forwarded actions and app metadata, so this information persists between browsing sessions
>It would be great to prune forwarded actions at some point (e.g. on completion, the target app can choose to dismiss it and remove it from its list of forwarded actions)

We'll set up a separate "archive" structure where executed events are stored, which isn't queried in the wrapper when actions are created or updated

>What happens if the forwarding path involves multiple forwarders? Right now, each forwarder (assuming they implement this API) would create a new entry in the forwarding actions, and the target app would likely be confused with the entries since there's nothing linking them together.

Because a forwarded action can only exist within one step in the path at a time, there are no foreseeable concurrency issues. We thought about a way to give each action a unique identifier, which would in turn allow us to link to prior steps in the path, but because there was no clear solution, we'll just reveal the remaining step (steps once multi-layered executionScripts can be passed into `decodeTransactionPath`).

>I am still thinking about the WebWorker script as being "committed" state, such that it should very closely mirror contract state. As such, my hunch is that the subscription to the forwarded actions should be a frontend-only use, likely masked in `@aragon/api-react`.

We found a solution where the store will subscribe to the `forwardedActions` observable in the section that holds state in "non-committed" memory as opposed to the cache

>I'm interested in exactly how you envision this being used. Is it to share some information between apps when a new event comes in (e.g. Projects to DotVoting)?

Yeah this will allow target apps to expose things like vote times in a current forwarder and/or deep linking urls. Long-term, we'd like to implement a pull-model where a forwarder will supply metadata upon request, but for this implementation Forwarders will push this info into the wrapper to be consumed by target apps

packages/aragon-api/src/index.js
topocount: We've decided that it's best to just pass this in every time, since we use the decoded "target" app as a key.

packages/aragon-wrapper/src/rpc/handlers/get-forwarded-actions.js
topocount: Because we're moving this endpoint into the store, should I leave this here now?

lkngtn:
This feature would help us provide a much better UX for Dandelion. Currently, we have a flow where someone submits a request to mint tokens, then a member of the organization must submit it to a vote, if the vote passes the request is approved, if it fails the request is rejected. Unfortunately, without the forwarder API we cannot provide users with any indication of the status of the request beyond that it is pending, approved, or withdrawn by the requestor. With the forwarder API we can update the status when the vote is submitted as well as if the request is rejected.

While it has not been merged we went ahead and implemented it for the token request app as a way to provide an additional data point and reference for how it would be used. 

The our implementation can be found in the forwader-api branch of the token request app:

https://github.com/1Hive/token-request-app/blob/7edbc591cfc14a26e491ea717282336442ee8ed9/app/src/script.js#L69

PR #317, opened 2019-05-29
App state caching: improvements++

sohkai:
Couple of changes / fixes, especially with race conditions.

I was getting a bit scared that we were flushing store updates without checking that they were first complete, since the IndexedDB API is all async (so that a second `cache()` request wouldn't have a race condition and override the previous one). I'm not certain if I _actually_ saw this happen, but there definitely was some funny business going on at times with the cache.

Also a good reminder for the future (to myself), the past events stream **only** ends when all the observables complete. This can sometimes make it look like an app's done syncing (it's showing all the relevant state), but actually is not yet (e.g. due to the past Vault events being slower than the Finance app's own events).

- 435c3a8: Likely still has a bit of a race with the frontend's first render, but at least flushes the old cache state immediately (rather than later, when past events start to stream)
- 7f3fe9e: Small off-by-one fix to not repeat blocks
- 3005248: See above; `store()` caching is now only done one at a time and `cache()` returns when the value has been committed
- cee9f85: See #315, `throttleTime()` is better than `sampleTime()`
- eb87d64: Updates the ACL caching strategy to be based on the `pastEvents()` as well. Particularly, some organizations (e.g. blankdao), have never touched their permissions after creation (I assume this to be true for > 90% of orgs), so the current ACL caching strategy doesn't actually cache anything.

### DIFF
--- a/packages/aragon-api/src/index.js
+++ b/packages/aragon-api/src/index.js
@@ -1,4 +1,4 @@
-import { forkJoin, from, merge } from 'rxjs'
+import { asyncScheduler, forkJoin, from, merge } from 'rxjs'
 import {
   catchError,
   delayWhen,
@@ -10,9 +10,9 @@ import {
   last,
   pluck,
   publishReplay,
-  sampleTime,
   startWith,
-  switchMap
+  switchMap,
+  throttleTime
 } from 'rxjs/operators'
 import Messenger, { providers } from '@aragon/rpc-messenger'
 
@@ -373,7 +373,8 @@ export class AppProxy {
         return getPastEvents(cachedBlock, pastEventsToBlock).pipe(
           mergeScan(wrappedReducer, { ...cachedState, ...initState }, 1),
           // throttle to reduce rendering and caching overthead
-          sampleTime(200),
+          // must keep trailing to avoid discarded events
+          throttleTime(1000, asyncScheduler, { leading: false, trailing: true }),
           delayWhen((state) => {
             console.debug('- store - reduced state from past event:', state)
             return this.cache('state', state)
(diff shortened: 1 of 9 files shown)


packages/aragon-api/src/index.js
```diff
@@ -346,7 +346,15 @@ export class AppProxy {
       map(v => v || {}))
     const latestBlock$ = this.web3Eth('getBlockNumber')
     // init the app state with the cached state
-    const initState$ = init ? cacheValue$.pipe(switchMap(({ state }) => from(init(state)))) : from([null])
+    const initState$ = init
+      ? cacheValue$.pipe(
+        switchMap(({ state }) => from(init(state))),
+        delayWhen((initState) => {
+          console.debug('- store - init state:', initState)
+          return this.cache('state', initState)
```
2color: This I assume ensures two things:
- The initialised state is persistent to cache
- The `forkJoin` will only take the value once `this.cache` is successful. 

Am I correct?

packages/aragon-wrapper/src/rpc/handlers/cache.js
```diff
@@ -7,8 +7,7 @@ export default function (request, proxy, wrapper) {
   }
 
   if (request.params[0] === 'set') {
-    wrapper.cache.set(cacheKey, request.params[2])
-    return Promise.resolve()
+    return wrapper.cache.set(cacheKey, request.params[2])
```
2color: Good catch!

packages/aragon-api/src/index.js
sohkai: Yes on both.

Unfortunately there's still a bit of a gap with the initialized state being flushed to `state` and the frontend rendering, but this will be taken care of (and we can remove this bit) once we [stop using the cache for `state`](https://github.com/aragon/aragon.js/pull/313#issuecomment-496577875)

PR #318, opened 2019-05-29
Use indexOf instead of regex

AquiGorka:


packages/aragon-wrapper/src/identity/LocalIdentityProvider.js
```diff
@@ -53,26 +53,20 @@ export default class LocalIdentityProvider extends AddressIdentityProvider {
...
-    const results = []
-
-    for (const [address, { name }] of Object.entries(identities)) {
-      if (isAddressSearch) {
-        searchRegex.test(address) && results.push({ name, address })
-      } else {
-        searchRegex.test(name) && results.push({ name, address })
-      }
-    }
+    const results = Array.from(Object.entries(identities))
```
2color: Is `Array.from` needed here?

PR #321, opened 2019-06-04
API to approve ERC20 on a different contract

### LINKED ISSUE #309: API to approve ERC20 on a different contract / context
We have a use case with the Aragon Fundraising [Apiary] app where we would need to use the `token` pretransaction hook to approve another contract than the app one.

This is because with the fundraising app users interact with a `Controller` or `API` contract itself relying on other contracts [such as the `MarketMaker` contract]. The contract which will actually move the token is not the `API` one - the one users interact with - but the `MarketMaker` one. So we would need to approve the `MarketMaker` contract from the `API` frontend ...

We can not release the Fundraising app without this feature so .... I would say it's urgent ;)

2color:
## Changes

- Allow passing an extra `spender` param when making an ERC20 token transaction to approve a different contract to the apps.
Fixes #309 

## Example:
To test this, set the `spender` to an  address in the Finance app's `intentParams.token` object:
https://github.com/aragon/aragon-apps/blob/7d61235044509095db09cf354f38422f0778d4bb/apps/finance/app/src/App.js#L58-L60


osarrouy:
It looks all good to me [though I can't officially 'review' the PR cause I dont' have the permission to do so].

packages/aragon-wrapper/src/utils/transactions.js
```diff
@@ -62,11 +62,13 @@ export async function createDirectTransaction (sender, app, methodName, params,
         console.warn(`${sender} already approved ${destination}. In some tokens, approval will fail unless the allowance is reset to 0 before re-approving again.`)
       }
 
+      // Aprrove the app (destination) unless an spender is passed to approve a different contract
```
izqui:
```suggestion
      // Approve the app (destination) unless an spender is passed to approve a different contract
```

packages/aragon-wrapper/src/utils/transactions.js
```diff
@@ -40,7 +40,7 @@ export async function createDirectTransaction (sender, app, methodName, params,
   }
 
   if (transactionOptions.token) {
-    const { address: tokenAddress, value: tokenValue } = transactionOptions.token
+    const { address: tokenAddress, value: tokenValue, spender } = transactionOptions.token
```
izqui: We should add this to the documentation about the `token` parameter (we could also fix the link to the Finance app example): https://github.com/aragon/aragon.js/blob/master/docs/APP.md#aragonapp

luisivan:
@izqui can you expand on that? Does that mean that apps can now approve withdrawals from different accounts to the one that the user is logged in with?

izqui:
@luisivan Once a token allowance is granted, the spender has control over transferring the allowed amount to any other address without user intervention (unless the spender is a smart contract and implements some restrictions for when tokens can be pulled from a user's account).

At the moment, the token parameter for an intent only allows apps to request a token allowance from the user's account to the app's smart contract, so only the app's smart contract can get tokens from the user account. Therefore, changing the code that can get tokens from a user's account requires an on-chain upgrade of the contract.

With this change, an app's frontend can set any address as the spender when creating a token allowance, and a frontend upgrade can change this logic silently, setting another address as the spender.

I suggest that at least we show more information about the token approval in the signer panel, so users can see the address that will be able to get tokens from their account. This is the only info that we are showing right now about an intent performing an approval pretransaction:
2019-06-05 at 11 44 47 AM

osarrouy:
@luisivan Not it means a user can allow a contract which is not directly the app's underlying contract to `transferFrom()` ERC20 on behalf of the user. The good news is that Metamask knows when such a approval happens and makes the transaction pretty clear before you sign it :)

osarrouy:
@izqui Agree about making it clearer for the user. Especially since in our case the contract being `approved` is still an aragon app [just not the current frontend's underlying contract] so we could easi make this readable like:

```
This transaction will allow MarketMaker to transfer `XX tokens` on your behalf.
```

2color:
> I suggest that at least we show more information about the token approval in the signer panel, so users can see the address that will be able to get tokens from their account. This is the only info that we are showing right now about an intent performing an approval pretransaction:

I agree. I'll give that a go.

2color:
Added a message in https://github.com/aragon/aragon/pull/825

luisivan:
> I suggest that at least we show more information about the token approval in the signer panel, so users can see the address that will be able to get tokens from their account. This is the only info that we are showing right now about an intent performing an approval pretransaction:

Agree as well, thanks for the details

PR #322, opened 2019-06-07
Identities autocomplete provider remove check for search term length

AquiGorka:
As discussed here: https://github.com/aragon/aragon.js/pull/308#discussion_r291205441

### DIFF
--- a/packages/aragon-wrapper/src/identity/LocalIdentityProvider.test.js
+++ b/packages/aragon-wrapper/src/identity/LocalIdentityProvider.test.js
@@ -138,8 +138,10 @@ test.serial('search should return an array of results of freely matching identit
   ]
   // map of search terms to expected count and names
   const searchTermToExpectation = {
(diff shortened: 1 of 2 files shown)


packages/aragon-wrapper/src/identity/LocalIdentityProvider.test.js
```diff
@@ -138,8 +138,8 @@ test.serial('search should return an array of results of freely matching identit
   ]
   // map of search terms to expected count and names
   const searchTermToExpectation = {
-    'ne': { names: [] },
-    '0x': { names: [] },
+    'xn': { names: [] },
```
2color: I'd actually add one test for searching with two chars with results, to indicate we're not ruling that out on the provider level

AquiGorka: Done, added more tests on tests with 1 and 2 chars.

PR #324, opened 2019-06-13
Delete selected labels

delfipolito:


packages/aragon-wrapper/src/index.js
```diff
@@ -935,9 +935,25 @@ export default class Aragon {
...
+  
+  /**
+   * Clear selected local identities
+   *
+   * @return {Promise<Object>}
+   */
+  
+  removeSelectedLocalIdentities (selectedAddresses) {
+    let self = this;
+    selectedAddresses.forEach(function(address){
```
AquiGorka: May I suggest:

```JavaScript
for(const address of selectedAddresses) {
  await this.identityProviderRegistrar.get('local').remove(address);
}
```

packages/aragon-wrapper/src/index.js
```diff
@@ -935,9 +935,23 @@ export default class Aragon {
...
+  
   clearLocalIdentities () {
     return this.identityProviderRegistrar.get('local').clear()
   }
+  
+  
+  /**
+   * Clear selected local identities
+   *
+   * @return {Promise<Object>}
```
AquiGorka: Let's remove this too as the function does not return anything

packages/aragon-wrapper/src/index.js
```diff
@@ -935,9 +935,23 @@ export default class Aragon {
...
+   */
+  
+  removeSelectedLocalIdentities (selectedAddresses) {
```
AquiGorka: This will probably need to be `async` no?

packages/aragon-wrapper/src/index.js
AquiGorka: cc @2color whats the team convention here? If the function is `async` it will return a `Promise` no? what do we set here?

packages/aragon-wrapper/src/index.js
```diff
@@ -935,9 +935,23 @@ export default class Aragon {
    *
    * @return {Promise<void>}
    */
+
```
2color: This line can be removed

packages/aragon-wrapper/src/index.js
```diff
@@ -935,9 +935,23 @@ export default class Aragon {
...
     return this.identityProviderRegistrar.get('local').clear()
   }
+  
+  
+  /**
+   * Clear selected local identities
+   *
+   * @return {Promise<Object>}
+   */
+
```
2color: I'd remove this line too to have the doc block right above

packages/aragon-wrapper/src/index.js
2color: Yes. The function needs to be async.

I'd also rename it to just `removeLocalIdentities`

packages/aragon-wrapper/src/index.js
2color: I'm pretty sure we just write Promise if the value doesn't matter that much, i.e. `@return {Promise}`

I'd add a @param documenting the type of selectedAddresses, e.g.
https://github.com/aragon/aragon.js/blob/fcdb8700c2d6e104b8115499e198768db38e54f9/packages/aragon-wrapper/src/index.js#L1201

packages/aragon-wrapper/src/index.js
```diff
@@ -938,6 +938,18 @@ export default class Aragon {
...
     return this.identityProviderRegistrar.get('local').clear()
   }
+  
+  
+  /**
+   * Clear selected local identities
+   *
+   * @return {Promise<Object>}
+   */
+    async removeLocalIdentities (selectedAddresses) {
```
AquiGorka: Extra whitespace here

2color:
Congrats @delfipolito! Thanks for your first contribution 🎇

PR #326, opened 2019-06-20
@aragon/api-react: usePath()

bpierre:
Receive and update the local app path.

Example:

```jsx
function App() {

  const { path, requestPath } = useAragonApi()

  // “Hello World” screen
  if (path === '/hello-world') {
    return (
      <div>
        <h1>Hello World</h1>
        <button onClick={() => requestPath('/')}>
          Back
        </button>
      </div>
    )
  }

  // Home
  return (
    <div>
      <button onClick={() => requestPath('/hello-world')}>
        Click
      </button>
    </div>
  )
}
```


packages/aragon-api-react/src/index.js
```diff
@@ -51,13 +56,19 @@ function AragonApi({
 
     let subscribers = []
 
-    const handleMessage = ({ data }) => {
-      if (data.from !== 'wrapper') {
+    const handleMessage = ({ data } = {}) => {
```
AquiGorka:
```suggestion
    const handleMessage = ({ data } = { from = null }) => {
```

packages/aragon-api-react/src/index.js
```diff
@@ -15,6 +15,10 @@ const requestMenu = () => {
   postMessage('requestMenu', true)
 }
 
+const requestPath = (path, appId) => {
+  postMessage('requestPath', [path, appId])
```
2color: How come we use postMessage for this rather than the rpc abstraction provided by `@aragon/api`?

packages/aragon-api-react/README.md
```diff
@@ -93,6 +93,41 @@ function App() {
 }
 ```
 
+#### `path` / `api.requestPath()`
+
+The app's current path. Its value is `"/"` by default.
+
+Use `api.requestPath()` to request the app be navigated to another path. Note that the navigation request _may_ be rejected, and in that case the `path` will stay constant.
```
bpierre: @sohkai Perhaps it was a bad idea, but up to this point the intention was to separate everything related to the client in different hooks, and expose `api` for either advanced features or contract-related APIs (intents, `.call`, `.events`, `.external`…). I was also thinking that we could continue splitting things out of `api` as the library progress, until RxJS is not exposed at all in this library. Of course `api` will stay accessible, so that authors always have the choice to interact with it directly if they prefer to do so.

`requestPath()` can be accessed directly through `usePath()`, but I think we should also expose it in `useAragonApi()`, and recommend to consume it this way in the documentation, to make it more like the rest. What do you think?

sohkai: Sounds good! I had no strong feelings on this either way. 

I'm not sure about being able to completely remove `api` from developers' minds, since the biggest thing is `external()` and intents, but perhaps we'll be able to encapsulate those concepts better in another major version of the API. Definitely the Proxy-ed calls for intents were a historical mistake.

In any case, AFAIK we don't expose any RxJS from api-react (outside of `api`'s methods)?

bpierre: > In any case, AFAIK we don't expose any RxJS from api-react (outside of api's methods)?

No we don’t, only in `api` :+1:

PR #327, opened 2019-06-21
@aragon/api-react: useApps()

bpierre:
Builds on https://github.com/aragon/aragon.js/pull/326

Receive the complete list of apps installed in the organization.

Example:

```jsx
import { useApps } from '@aragon/api-react'
function App() {
  const apps = useApps()
  return (
    <div>
      {apps.map(app => (
        <div>{app.appAddress}</div>
      ))}
    </div>
  )
}
```

### DIFF
--- a/packages/aragon-api-react/README.md
+++ b/packages/aragon-api-react/README.md
@@ -151,6 +151,22 @@ function App() {
 }
 ```
 
+#### `currentApp`
+
+Details about the current app. It returns a single object with the same keys as the objects in
+`apps`.
+
+Example:
+
+```jsx
+function App() {
+  const { currentApp } = useAragonApi()
+  return (
+    <div>{currentApp.appAddress}</div>
+  )
+}
+```
+
 #### `network`
 
 An [object](https://github.com/aragon/aragon.js/blob/master/docs/API.md#network) representing the current network using its `id` and `type` entries. Its value is `null` until it gets loaded.
--- a/packages/aragon-api-react/README.md
+++ b/packages/aragon-api-react/README.md
@@ -108,36 +108,6 @@ function App() {
 }
 ```
 
-#### `apps`
-
-The complete list of apps installed in the organization. Its value is an empty array (`[]`) until
-the list of apps are loaded.
-
-Each object in the array holds the following keys:
-
-- `appAddress`: the app's contract address
-- `appId`: the app's appId
-- `appImplementationAddress`: the app's implementation contract, if any (only available if this app is a proxied AragonApp)
-- `identifier`: the app's identifier, if any
-- `isForwarder`: whether the app is a forwarder or not
-- `kernelAddress`: the kernel address of the organization this app is installed on (always the same)
-- `name`: the app's name, if available
-
-Example:
-
-```jsx
-function App() {
-  const { apps } = useAragonApi()
-  return (
-    <div>
-      {apps.map(app => (
-        <div>{app.appAddress}</div>
-      ))}
-    </div>
-  )
-}
-```
-
 #### `connectedAccount`
 
 The connected Ethereum account. Its value is `""` (empty string) when there is no account connected.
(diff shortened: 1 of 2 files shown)


packages/aragon-api-react/src/index.js
```diff
@@ -122,27 +141,32 @@ function getAragonApiData(hookName) {
...
 const useApi = () => getAragonApiData('useApi()').api
 const useAppState = () => getAragonApiData('useAppState()').appState
+const useApps = () => getAragonApiData('useApps()').apps
 const useConnectedAccount = () =>
   getAragonApiData('useConnectedAccount()').connectedAccount
-const useMenuButton = () => {
-  const { displayMenuButton, requestMenu } = getAragonApiData('useMenuButton()')
-  return [displayMenuButton, requestMenu]
-}
+const useMenuButton = () => [
```
2color: is this an API change as it returns an array tuple now?

packages/aragon-api-react/src/index.js
```diff
@@ -27,6 +31,8 @@ function AragonApi({
   const [network, setNetwork] = useState(null)
   const [appState, setAppState] = useState(null)
   const [displayMenuButton, setDisplayMenuButton] = useState(false)
+  const [path, setPath] = useState('/')
+  const [apps, setApps] = useState([])
```
2color: is `apps` an array?

packages/aragon-api-react/src/index.js
```diff
@@ -51,13 +57,23 @@ function AragonApi({
...
+
       if (data.name === 'displayMenuButton') {
         setDisplayMenuButton(data.value)
       }
+
+      if (data.name === 'path' && data.value !== path) {
+        setPath(data.value)
+      }
+
+      if (data.name === 'apps' && Array.isArray(data.value)) {
```
2color: Is this message already being sent?

sohkai:
@bpierre Pushed 3de8d3b to update this implementation with `@aragon/api`'s. Perhaps it may make sense to rebase the `apps` changes on master and merge it separately from path?

packages/aragon-api-react/src/index.js
bpierre: I don’t think there was any change here? It’s been returning an array since the [initial PR](https://github.com/aragon/aragon.js/pull/259/files#diff-df3ed8cb308cbe434aac1d4a00e05543R139).

packages/aragon-api-react/src/index.js
bpierre: Yes :-)

packages/aragon-api-react/src/index.js
bpierre: This part has been replaced by `api.getApps()`.

packages/aragon-api-react/README.md
```diff
@@ -108,6 +108,36 @@ function App() {
...
 ```
 
+#### `apps`
+
+The complete list of apps installed in the organization. Its value is an empty array (`[]`) until
+the list of apps are loaded.
+
+Each object in the array holds the following keys:
+
+- `appAddress`: the app's contract address
```
bpierre: @sohkai Is the idea to progressively move away from `proxyAddress` / `contractAddress`? And shouldn’t we pass `currentApp` through `transformAppInformation(app)` too?

sohkai: > And shouldn’t we pass currentApp through transformAppInformation(app) too?

I think we do this: https://github.com/aragon/aragon.js/blob/master/packages/aragon-wrapper/src/rpc/handlers/get-apps.js#L53

> Is the idea to progressively move away from proxyAddress / contractAddress?

I'd like to. `proxyAddress` has always been a bit of a misnomer, and confusing. `contractAddress` even more so. The important thing, outside of aragonOS (and some tiny parts in aragon.js), is that each app has a contract address. Maybe it's a proxy, maybe it has a base implementation, but these details shouldn't matter to any developer unless they're building something custom on the contract level (at which point aragon.js probably isn't the library for them).

bpierre: > I think we do this: https://github.com/aragon/aragon.js/blob/master/packages/aragon-wrapper/src/rpc/handlers/get-apps.js#L53

Oh I missed that :+1: 

> I'd like to. proxyAddress has always been a bit of a misnomer, and confusing. contractAddress even more so. The important thing, outside of aragonOS (and some tiny parts in aragon.js), is that each app has a contract address. Maybe it's a proxy, maybe it has a base implementation, but these details shouldn't matter to any developer unless they're building something custom on the contract level (at which point aragon.js probably isn't the library for them).

That sounds good yes! I wonder if we should also move to this naming in aragon/aragon, just for consistency.

sohkai: Haha absolutely... we can start by mapping the values over to a different schema in aragonjs-wrapper.

It's a bit hard to update `@aragon/wrapper` because the CLI also depends on it, but I hope with the refactor they're doing they'll be able to accommodate for breaking changes more easily.

PR #328, opened 2019-06-25
Feat/external txs

### LINKED ISSUE #120: Add ability to handle external transaction intents

Schwartz10:
**Changes**

Fixes #120


Schwartz10:
Hey! You'll notice I added a `getApps` method that the front-end clients can use to get the proxy address of external apps. This might be redundant with https://github.com/aragon/aragon.js/pull/327 so please let me know if I should remove. Thanks!

topocount:
It'd be awesome if we could interact with contracts that aren't registered as a DAO's app contracts. The primary use case for this right now is interacting with [StandardBounties.sol](https://github.com/Bounties-Network/StandardBounties/blob/develop/contracts/StandardBounties.sol), Users of the current Rinkeby deployment need to relay their calls through the projects contract. This adds what would otherwise be unnecessary expense, and isn't the cleanest way to go about this.

Schwartz10:
> It'd be awesome if we could interact with contracts that aren't registered as a DAO's app contracts.

@topocount let me know if this is not working for you, and if so, what the error message is.

sohkai:
> It'd be awesome if we could interact with contracts that aren't registered as a DAO's app contracts.

@topocount Agreed, @Schwartz10 and I discussed this yesterday about using the transaction pathing only if the transaction is to an installed app. Otherwise it'll be treated as a "direct" transaction.

Schwartz10:
@sohkai @2color ready for your review! 

One thing to note is that the transaction path contains a transaction object that now has 2 additional fields: `installedApp` and `external`

`installedApp && external` means an external intent on another application contract within your DAO
`!installedApp && external` means an external intent on a completely isolated contract (like @topocount use case with standard bounties)

Figured it's probably useful to give this information in the client, but I couldn't figure out a better way to display that information. This is what the tests I added are essentially checking for

docs/API.md
```diff
@@ -168,18 +168,15 @@ Returns **[Observable](https://rxjs-dev.firebaseapp.com/api/index/class/Observab
 
 Creates a handle to interact with an external contract (i.e. a contract that is **not** your app's smart contract, such as a token).
 
-> **Note**<br>
-> Sending transactions to these external contracts is not yet supported as additional security and disclosure enhancements are required in frontend clients (this is a large attack vector for malicious applications to invoke dangerous functionality).
-
 #### Parameters
 
-- `address` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)**: The address of the external contract
+- `address` **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)**: The proxy address of the external contract
```
sohkai: Note that this is not necessarily true, especially for `external` (which may be anything like tokens or ENS or etc.). They do not have to be proxy contracts.

packages/aragon-wrapper/src/rpc/handlers/external.js
```diff
@@ -16,6 +16,35 @@ export function call (request, proxy, wrapper) {
...
+      proxyAddress,
+      method,
+      params
+    )
+
+  return wrapper.performTransactionPath(
+    transactionPath.map((tx) => ({
+      ...tx,
+      external: true,
+      installedApp: !!installedApp
```
sohkai: As mentioned in https://github.com/aragon/aragon/pull/850#discussion_r299933995, we don't _really_ need this, since we can check in the client if the `to` is an installed app or not.

packages/aragon-wrapper/src/index.js
```diff
@@ -1319,6 +1320,42 @@ export default class Aragon {
...
+  /**
+   * Calculate the transaction path for a transaction to `destination`
+   * that invokes `methodName` with `params`.
+   *
+   * @param  {string} destination An address of an uninstalled smart contract application
+   * @param  {string} method ABI of smart contract method
+   * @param  {Array<*>} params
+   * @return {Promise<Array<Object>>} An array of a single Ethereum direct transaction
+   */
+  async getUninstalledAppTransactionPath (destination, method, params) {
```
sohkai: Instead of having two functions, we could probably just combine this with `getTransactionPath()` and handle it as a case where the `destination` is not an installed app.

Schwartz10:
Thanks for the feedback @sohkai ! My goal is to wrap this up tomorrow.

packages/aragon-wrapper/src/rpc/handlers/external.js
Schwartz10: Are you talking just about the `installedApp` var or `external` too? It seems apps would need a way to determine their own proxy address to determine `external`, but `installedApp` could definitely be calculated easily.

sohkai: Just `installedApp`. `external` is a little bit harder to get because it requires some context about the currently running app (which the client does still have, but involves a bit more state plumbing).

packages/aragon-wrapper/src/index.js
Schwartz10: Hey @sohkai - originally I tried to approach external tx pathing this way, but what's the best strategy to handle the ABI of the external method without breaking things? 

One approach is to include an optional methodAbi param:

```js
 async getTransactionPath (destination, methodName, params, finalForwarder, methodAbi = []) { ... }
```

Another way that feels prettier, but also like a much bigger change, is to adjust the `getTransactionPath` method to take `method` (includes the abi) rather than just the `methodName`, and then grab the name from the `method` object. This is pretty much exactly what `getUninstalledAppTransactionPath` does: [method is used as abi](https://github.com/aragon/aragon.js/pull/328/files#diff-5b914b5b526a57f89974fa2fe009ee80R1338] & [using method.name as methodName](https://github.com/aragon/aragon.js/pull/328/files#diff-5b914b5b526a57f89974fa2fe009ee80R1344)

Although more concise, this strategy _feels_ like its changes could propagate to other parts of aragon.js and potentially break things. 

Am i missing something? or is there another strategy you like better that I didn't think of?

Schwartz10:
@sohkai i'm confident these changes will work, but i want to test them in some real code on monday. Feel free to review whenever is convenient for you, and I'll leave a comment here once I've tested thoroughly

Schwartz10:
@sohkai this is all set - tested with both installed, external contracts, and uninstall external contracts.

sohkai:
@Schwartz10 I've just added a few commits with mostly clarifications / small cleanups I saw. The big important one is 0efdaa4, where I've restructured a few of the internal plumbing. The only external API change is how `external` is propagated; now it's on the top-level object emitted by the `transactions` observable, rather than part of each transaction.

We still need to rework the external tests a bit. Here's what I'm thinking:

- In `rpc/handlers/external.test.js`, use a mocked `wrapper` and we just stub and spy that the right functions are being called with the right arguments
- In `index.test.js`, test out the flow, using most of the parts of your current test in `rpc/handlers/external.test.js`

It would be great if you could review and retest these changes!

sohkai:
@Schwartz10 I've pushed the final changes for the tests, including a bug fix in 31a6236 on my previous pushes.

It also changes the language around the "JSON interface". web3.js likes to use `jsonInterface` in its argument names but it's quite unclear what it means (the ABI array format, or the specific method / event object description in the array?), so I've fallen back to using `JsonDescription` for objects since that's what the [ABI-spec](https://solidity.readthedocs.io/en/latest/abi-spec.html#abi-json) uses.

It would be greatly appreciated if you could re-test, and we can get this merged!

Schwartz10:
@sohkai just re-tested with external installed, and external uninstalled. Both working properly. Let's pull this in asap! Once that's all set, I'll work on finishing up https://github.com/aragon/aragon/pull/850

Issue #194, opened 2018-10-24
Add ability for apps to get current list of apps


sohkai:
It'd be nice if apps had information about what apps were currently installed, as well as information about themselves (e.g. address and identifier).

bpierre:
> about what apps were currently installed

Do we want any app to know about the other apps that are currently installed? I’m wondering if this could have any security implications.

sohkai:
_Technically_ they would know anyway, since it's all public info they could get from the ACL on-chain. Even if other apps knew about each other, our security model should still be kept if we make sure to explicitly warn about any interaction between apps—if you try to do an action on an app but the transaction's actually to another contract / app, the Aragon client should warn you about it and make sure you know what you're doing.

PR #332, opened 2019-07-03
Feat/get apps

Schwartz10:
**Changes**

Adds a method to get the list of installed applications. @sohkai requested to add this functionality in a separate PR from https://github.com/aragon/aragon.js/pull/328

Fixes https://github.com/aragon/aragon.js/issues/194.


docs/APP.md
```diff
@@ -108,13 +108,19 @@ const frontendOfApp = new AragonApp(new providers.WindowMessage(window.parent))
 
 Get an array of the accounts the user currently controls over time.
 
-Returns **[Observable](https://rxjs-dev.firebaseapp.com/api/index/class/Observable)**: An multi-emission observable that emits an array of account addresses every time a change is detected.
+Returns **[Observable](https://rxjs-dev.firebaseapp.com/api/index/class/Observable)**: A multi-emission observable that emits an array of account addresses every time a change is detected.
```
Schwartz10: I snuck in this little grammar change :) But let me know if you want me to remove

bpierre:
Nice! Once merged I’ll update https://github.com/aragon/aragon.js/pull/327 to use it.

Schwartz10:
@sohkai thanks for adding some commits here! Multi-emission event should be good - we just need to get the list of all apps, find the Discussion App, and get its proxy address + abi. As long as we can still do that with multi-emission events (which we should), then it seems appropriate.

PR #333, opened 2019-07-03
Bump Web3 to  r37

topocount:
**Changes**
I bumped the web3 dependency up to release 37 from release 33. I've tested and verified that this will support the ability to return tuples from contracts without any further modifications.


topocount:
Closing  because I opened #334 after rebasing changes to aragon/master

PR #334, opened 2019-07-03
Web3: upgrade to latest v1

topocount:
**Changes**
I bumped the web3 dependency up to release 37 from release 33. I've tested and verified that this will support the ability to return tuples from contracts without any further modifications.

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -12,7 +12,7 @@ module.exports = [
   {
     name: '@aragon/wrapper',
     path: "packages/aragon-wrapper/dist/index.js",
-    limit: "850 KB"
+    limit: "475 KB"
   },
   {
     name: '@aragon/api-react',
--- a/packages/aragon-wrapper/package.json
+++ b/packages/aragon-wrapper/package.json
@@ -82,7 +82,7 @@
     "js-sha3": "^0.7.0",
     "localforage": "^1.7.3",
     "localforage-memoryStorageDriver": "^0.9.2",
-    "radspec": "^1.3.0",
+    "radspec": "^1.4.0",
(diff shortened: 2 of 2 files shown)


topocount:
@sohkai do you know of any workarounds for the increased size of this web3 release? Or will we have to bump the size limit?

sohkai:
@topocount We'll likely just up the size limit; there's not much we can do here.

Before this gets merged I'd like to test it (and up all the pinned versions of web3.js _everywhere_ to beta.37).

topocount:
> Before this gets merged I'd like to test it (and up all the pinned versions of web3.js _everywhere_ to beta.37).

I tested this PR, along with a version of the client running web3-1.0r37 with the TPS test script and didn't run into any issues. However, while our apps do use a fair bit of the client and wrapper's functionality, not all of it was tested there.

sohkai:
It occurred to me that it's likely the additional size is because we've pinned web3.js to other versions in `@aragon/wrapper`'s dependencies (likely `@aragon/apm`, which I'd like to remove anyway). This will likely not be a problem once we bump those as well.

topocount:
> It occurred to me that it's likely the additional size is because we've pinned web3.js to other versions in `@aragon/wrapper`'s dependencies (likely `@aragon/apm`, which I'd like to remove anyway). This will likely not be a problem once we bump those as well.

@sohkai aside from what's quoted above, is there anything remaining you'd need me to do to get this merged?

sohkai:
@topocount Nope, it’s been mostly on me; I want to remove the `apm.js` dependency this week and then I’ll have this all tested and upgraded through the stack.

PR #337, opened 2019-07-08
Transactions: Decode nested forwarding paths

facuspagnuolo:
This PR aims to improve the way we decode transactions path to support nested transaction paths. 
This will help describing complex chained forwarding asynchronous actions, for example, a voting proposal that will trigger another voting proposal to finally execute sth in case it is successful.


packages/aragon-wrapper/src/evmscript/index.js
```diff
@@ -1,5 +1,6 @@
 import abi from 'web3-eth-abi'
 
+export const FORWARD_SIG = '0xd948d468'
```
sohkai:
```suggestion
export const FORWARD_SIG = '0xd948d468' // function forward(bytes)
```

packages/aragon-wrapper/src/index.js
```diff
@@ -1528,6 +1530,22 @@ export default class Aragon {
...
+        const forwardDataStartIndex = forwardDataOffset + 64
+        const forwardDataLength = parseInt(`0x${forwardCallData.substring(forwardDataOffset, forwardDataStartIndex)}`, 16) * 2
+        const forwardData = `0x${forwardCallData.substring(forwardDataStartIndex, forwardDataStartIndex + forwardDataLength)}`
+
+        // If forwarding data is a script decode it
+        const forwardScriptId = script.substring(0, 10)
+        if (forwardScriptId === CALLSCRIPT_ID) {
+          segment.children = this.decodeTransactionPath(forwardData)
+        }
+      }
```
sohkai: Let's add some comments here to explain the offsetting. Best to be explicit (I get confused all the time looking at offsets). Perhaps "munching" on each operation (i.e. consuming the first part of the string via `substring(x)`) would also make it more clear.

We may also want to add in a bit of error checking for the lengths in case someone puts in garbage data into an EVMScript.

packages/aragon-wrapper/src/index.js
```diff
@@ -1540,7 +1558,7 @@ export default class Aragon {
    * @param  {Array<Object>} path
    * @return {Promise<Array<Object>>} The given `path`, with decorated with descriptions at each step
    */
-  describeTransactionPath (path) {
+  async describeTransactionPath (path) {
```
sohkai: I think this can be left non-async, since it already returns a promise

packages/aragon-wrapper/src/index.test.js
```diff
@@ -1374,13 +1375,75 @@ test('should be able to decode an evm call script with multiple transactions', a
...
+    data: '0x' +
+          'd948d468' +                                                            // forward signature
+          '0000000000000000000000000000000000000000000000000000000000000020' +    // offset
+          '0000000000000000000000000000000000000000000000000000000000000060' +    // 96 data length
+          '00000001' +                                                            // spec id
+          '14a3208711873b6aab2005f6cca0f91658e287ef' +                            // forward target
+          '00000044' +                                                            // 68 data length
+          '9dc29fac' +                                                            // burn
+          '0000000000000000000000008401eb5ff34cc943f096a32ef3d5113febe8d4eb' +    // token holder
+          '0000000000000000000000000000000000000000000000000de0b6b3a7640000'      // 1e18
```
sohkai: Would be nice to use another method that has a different argument length instead, e.g. `Finance.transfer()`

packages/aragon-wrapper/src/index.js
```diff
@@ -1528,6 +1530,22 @@ export default class Aragon {
...
+
+      if (selector === FORWARD_SIG) {
+        const forwardCallData = data.substring(10)
+        const forwardDataOffset = parseInt(`0x${forwardCallData.substring(0, 64)}`, 16) * 2
+        const forwardDataStartIndex = forwardDataOffset + 64
+        const forwardDataLength = parseInt(`0x${forwardCallData.substring(forwardDataOffset, forwardDataStartIndex)}`, 16) * 2
+        const forwardData = `0x${forwardCallData.substring(forwardDataStartIndex, forwardDataStartIndex + forwardDataLength)}`
+
+        // If forwarding data is a script decode it
+        const forwardScriptId = script.substring(0, 10)
```
sohkai:
```suggestion
        const forwardScriptId = forwardData.substring(0, 10)
```

I think we should be using the forwarded script here?

facuspagnuolo: nice catch haha, it was working by chance

packages/aragon-wrapper/src/index.js
facuspagnuolo: true, but sometimes it's useful to tell if the the functions is actually returning a promise or not from its definition adding the `async` keyword, otherwise you will need to read the function to tell if that's the case... but I can remove it if we already decided to do the other way around :)

sohkai: No real preference, and the code styling on aragon.js is a bit awkward compared to the other repos we have (to be fixed soon). Generally I think we avoid `async` unless there's an explicit `await` :shrug:.

facuspagnuolo:
Thanks @sohkai for reviewing! I already addressed your comments if you wanna take another look 🙌

PR #342, opened 2019-07-10
Wrapper: cosmetic changes to removeLocalIdentities()

sohkai:


packages/aragon-wrapper/src/index.js
```diff
@@ -957,12 +957,13 @@ export default class Aragon {
   /**
    * Remove selected local identities
    *
-   * @param {Array<Address>} addresses The addresses to be removed from the local identity provider
+   * @param {Array<string>} addresses The addresses to be removed from the local identity provider
```
sohkai: I think we generally keep addresses to a `string` type in the documentation.

PR #344, opened 2019-07-10
Wrapper: add ability to apply token approval pretransactions to forwarders requiring fees

sohkai:
Adds the ability to attach pretransactions to an initial forwarder that requires fees (e.g. a [Tollgate](https://github.com/aragonone/tollgate)).

Requires a new release of aragonOS (https://github.com/aragon/aragonOS/pull/539).

### DIFF
--- a/packages/aragon-wrapper/package.json
+++ b/packages/aragon-wrapper/package.json
@@ -74,7 +74,7 @@
   "dependencies": {
     "@aragon/apm": "^3.0.0",
     "@aragon/rpc-messenger": "^2.0.0-beta.1",
-    "@aragon/os": "^4.0.1",
+    "@aragon/os": "^4.2.1",
     "@babel/runtime": "^7.1.2",
     "dot-prop": "^4.2.0",
     "eth-ens-namehash": "^2.0.8",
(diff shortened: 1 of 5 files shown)


packages/aragon-wrapper/src/utils/transactions.js
```diff
@@ -39,44 +116,8 @@ export async function createDirectTransaction (sender, app, methodName, params,
...
-
-    // If allowance is already greater than or equal to amount, there is no need to do an approve transaction
-    if (allowanceBN.lt(tokenValueBN)) {
-      if (allowanceBN.gt(toBN(0))) {
-        // TODO: Actually handle existing approvals (some tokens fail when the current allowance is not 0)
-        console.warn(`${sender} already approved ${destination}. In some tokens, approval will fail unless the allowance is reset to 0 before re-approving again.`)
-      }
-
-      // Approve the app (destination) unless an spender is passed to approve a different contract
-      const approveSpender = spender || destination
```
sohkai: Oops, this was actually done a bit too late; we should've used the `spender` when checking the initial allowance.

packages/aragon-wrapper/package.json
```diff
@@ -74,7 +74,7 @@
   "dependencies": {
     "@aragon/apm": "^3.0.0",
     "@aragon/rpc-messenger": "^2.0.0-beta.1",
-    "@aragon/os": "^4.0.1",
+    "@aragon/os": "^4.2.1",
```
2color: `4.2.1` hasn't been released yet

sohkai: Just released :).

PR #346, opened 2019-07-15
Wrapper: fix no path found for intent baskets

sohkai:
Small fix for when no transaction path could be found for an intent basket.

Fixes the crash that occurs when trying to upgrade all of the [AGP org's](https://mainnet.aragon.org/#/governance.aragonproject.eth/) apps cc @izqui.

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -1336,15 +1336,11 @@ export default class Aragon {
    * @param  {Object} [options]
    * @param  {string} [options.checkMode] Path checking mode
    * @return {Promise<Object>} An object containing:
-   *   - `direct` (boolean): whether the current account can directly invoke this basket
-   *     (requiring separate transactions)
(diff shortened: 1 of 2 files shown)


packages/aragon-wrapper/src/index.js
```diff
@@ -1416,7 +1416,7 @@ export default class Aragon {
     // Failed to find a path
     return {
       direct: false,
-      transactions: []
+      path: []
```
sohkai: Maybe we should just return both arrays all the time; right now the API is:

- `direct: false`: includes `path`
- `direct: true`: includes `transactions`

sohkai: Done in 28099da; I've opted to ensure both arrays are returned and removed `direct`, as it no longer really serves a purpose (it never really did, and just added API bloat). Users can tell whether the intent basket is possible or not by checking if either of `path`, or `transactions` is non-empty.

This is a breaking change with an associated fix in https://github.com/aragon/aragon/pull/886.

PR #348, opened 2019-07-17
Configurable cache driver

AquiGorka:


packages/aragon-wrapper/src/index.js
```diff
@@ -151,7 +151,8 @@ export default class Aragon {
     const defaultOptions = {
       apm: {},
       defaultGasPriceFn: () => { },
-      provider: detectProvider()
+      provider: detectProvider(),
+      cacheDriver: null
```
sohkai: My main hesitation around this is what a proper API should look like, and I'm starting to think something like:

```js
cache: {
  forceLocalStorage: true
}
```

Might be enough. I assume nobody wants to use WebSQL (let's just remove it from the default list), and I don't know anything about its security implications (now that we know IndexedDB is a PITA), so let's just give users the ability to "force" the localstorage downgrade if they want to.

packages/aragon-wrapper/src/index.js
```diff
@@ -165,7 +166,7 @@ export default class Aragon {
     this.kernelProxy = makeProxy(daoAddress, 'Kernel', this.web3)
 
     // Set up cache
-    this.cache = new Cache(daoAddress)
+    this.cache = new Cache(daoAddress, options)
```
```suggestion
    this.cache = new Cache(daoAddress, driver)
```

AquiGorka:
@sohkai and @2color updated with your suggestions :)

PR #350, opened 2019-07-24
Wrapper: add global configuration store for cache and event options

sohkai:
**Adds** a simple global configuration store, based on the `Aragon` class' given options. Note that this does not work well if you create multiple `Aragon` classes and use different options, but let's not worry about that for now.

**Adds** a global option to delay new events from event subscriptions (`eth_subscribe`). This is necessary when the connected ETH node (i.e. Infura) has race conditions between sending events and commiting internal state, such that sometimes an event from `eth_subscribe` will come before its transaction can be reflected from `eth_call`s. Adding a small delay between receiving the event to processing it gives the node some time to consolidate its internal state.

Also shuffles a few things around, in a backwards-compatible way:

- Moves the `forceLocalStorage` option to be global, as it should affect every cache being created cc @AquiGorka 
- Changes the internal API for `ContractProxy` to take in explicit options for less-widely used arguments


packages/aragon-wrapper/src/configuration/index.js
```diff
@@ -0,0 +1,11 @@
+// Simple key-value configuration store
+const CONFIGURATION = {
+}
+
+export function getConfiguration (key) {
+  return key ? CONFIGURATION[key] : CONFIGURATION
```
AquiGorka: Should this a) check if the key exists? and return null if it does not?

sohkai: I think it's probably fine just to return `undefined` if the configuration doesn't exist 🤷‍♂

AquiGorka: ja! didn't catch that so yeah!

PR #352, opened 2019-07-25
API: add path support (navigation)

ottodevs:
**Changes**
Added request handler to the wrapper, then handle path from there and finally connect API and slightly modify API-react to use this new handler from the observable

Pending docs until some approval feedback / polish

TBD

### DIFF
--- a/packages/aragon-api/src/index.js
+++ b/packages/aragon-api/src/index.js
@@ -131,7 +131,7 @@ export class AppProxy {
    *
    * @return {Observable} Multi-emission Observable that emits the app's current path every time a change is detected.
    */
-  paths () {
+  path () {
     return this.rpc.sendAndObserveResponses(
       'path',
       ['observe']
--- a/packages/aragon-api/src/index.js
+++ b/packages/aragon-api/src/index.js
@@ -61,19 +61,6 @@ export class AppProxy {
   }
 
   /**
-   * Get current path.
-   *
-   * @return {Observable} Multi-emission Observable that emits a path every time a change is detected.
-   */
-  path () {
-    return this.rpc.sendAndObserveResponses(
-      'path'
-    ).pipe(
-      pluck('result')
-    )
-  }
-
-  /**
    * Get the network the app is connected to over time.
    *
    * @return {Observable} Multi-emission Observable that emits an object with the connected network's id and type every time the network changes.
@@ -140,6 +127,34 @@ export class AppProxy {
   }
 
   /**
+   * Get current path for the app. Useful for in-app routing and navigation.
+   *
+   * @return {Observable} Multi-emission Observable that emits the app's current path every time a change is detected.
+   */
+  paths () {
+    return this.rpc.sendAndObserveResponses(
+      'path',
+      ['observe']
+    ).pipe(
+      pluck('result')
+    )
+  }
+
+  /**
+   * Request a new path.
+   *
+   * @return {Observable} Single-emission Observable that emits if the path request succeeded and errors if rejected
+   */
+  requestPath (path) {
+    return this.rpc.sendAndObserveResponse(
+      'path',
+      ['modify', path]
+    ).pipe(
+      pluck('result')
+    )
+  }
+
+  /**
    * Resolve an address' identity, using the highest priority provider.
    *
    * @param  {string} address Address to resolve.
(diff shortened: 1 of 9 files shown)


packages/aragon-api-react/src/index.js
```diff
@@ -81,6 +77,9 @@ function AragonApi({
 
           // network
           api.network().subscribe(network => setNetwork(network || null)),
+
+          // path
+          api.path().subscribe(path => setPath(path)),
```
sohkai: Does a default value make sense here? Like a `''` or `'/'`?

packages/aragon-wrapper/src/rpc/handlers/path.js
```diff
@@ -0,0 +1,3 @@
+export default function (wrapper) {
+    return wrapper.path
+}
```
sohkai: It probably makes sense to use a similar pattern to the [`cache`](https://github.com/aragon/aragon.js/blob/master/packages/aragon-wrapper/src/rpc/handlers/cache.js#L3) handler here (`get` vs. `request` path).

packages/aragon-api/src/index.js
```diff
@@ -61,6 +61,19 @@ export class AppProxy {
...
+   * @return {Observable} Multi-emission Observable that emits a path every time a change is detected.
+   */
+  path () {
+    return this.rpc.sendAndObserveResponses(
+      'path'
+    ).pipe(
+      pluck('result')
+    )
+  }
+
```
sohkai: Reminder to add `requestPath()` here and in `api-react` :)

packages/aragon-wrapper/src/index.js
```diff
@@ -988,6 +989,27 @@ export default class Aragon {
...
   }
 
+  /**
+   * Initialise the path observable.
+   *
+   * @return {Promise<void>}
+   */
+  async initPath () {
+    this.path = new ReplaySubject(1)
+    this.path.next('/')
```
sohkai: This observable is the only one that I'd think a bit more on, and be careful with. A couple of thoughts:
- Having `this.path` transparently represent the current app isn't safe if a script wanted to try and use `requestPath()`.
- What we probably want is a request/response scheme, where there's an `this.pathIntents` (for sake of consistency with other "intents") that the client receives requests on, and it either accepts or rejects those intents.
- If a path change intent is accepted, we push a change to `this.path` with the proxy address and the accepted path. At any time, the client can use `setPath(proxyAddress, path)` to change the path for an app (useful, for example, when loading the org at `<org>/<voting address>/vote/3`; the client should immediately tell the voting app to load the vote once the handshake is complete).
- The `path` handler should filter out `this.path` emissions that don't match its own proxy address

sohkai:
@ottodevs I've done a few things to this PR:

- Rebased it against master
- Cherry-picked `api-react` changes to that PR (#326) 
- Added request path APIs
- Updated the implementation based on https://github.com/aragon/aragon.js/pull/352#discussion_r307400300

In particular, clients should now respond to `pathIntent`s and set the corresponding app's path using `wrapper.setAppPath()` when allowing a request.

packages/aragon-api/src/index.js
```diff
@@ -139,6 +126,34 @@ export class AppProxy {
     )
   }
 
+  /**
+   * Get current path for the app. Useful for in-app routing and navigation.
+   *
+   * @return {Observable} Multi-emission Observable that emits the app's current path every time a change is detected.
+   */
+  paths () {
```
bpierre: Shouldn’t it be named `path`? Even if it updates, it’s always a single path that is being received.

sohkai: Hmm yes, we do name `network` as singular even though you're observing the network changes over time. Let's rename it back to `path()`.

`getApps()` should be named `apps` or `installedApps`, but is a bit problematic due to backwards compatibility issues. I'll add `installedApps` and deprecate `getApps()`.

sohkai:
@bpierre I've also thought about if it made sense to pass along more than a string; perhaps we can pass a string and an optional object that contains client-specific behaviour.

Ideally asking to navigate to something like the email notifications would be more robust than harding coding the client's preferences query string into the path request.

But we can do this on a second iteration since this change would be backwards compatible with the current API.

packages/aragon-api/src/index.js
sohkai: Changes for `installedApps()` and `currentPath()` are in https://github.com/aragon/aragon.js/pull/374.

`paths()` has been renamed to `path()` in 1e251f1.

PR #360, opened 2019-08-16
Add web3Eth examples

0xGabi:
**Changes**

When I helped @rperez89 do some work on a new aragon app was difficult to understand how the `web3Eth` function call worked, especially cause the official documentation use it in a different way. I have included a couple of examples to further explain it

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -269,7 +269,7 @@ api.web3Eth('getTransactionReceipt', trxHash).subscribe(
 ```
 
 ```javascript
-api.web3Eth('getBlock', blockNumber).first()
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -255,6 +255,27 @@ Currently the white-list includes:
...
+    // use receipt
+  },
+  err => {
+    // handle error
+  }
+)
+```
+
+```javascript
+api.web3Eth('getBlock', blockNumber).first()
```
sohkai:
```suggestion
const block = api.web3Eth('getBlock', blockNumber).toPromise()
```

PR #361, opened 2019-08-20
API: event trigger (#361)

### LINKED ISSUE #119: Add ability for frontend to trigger events in the script

topocount:
**Changes**

This submission resolves #119 and restores the ability to communicate with the script reducer from the frontend, but in an officially documented fashion.

- [X] I have updated the associated documentation with my changes


osarrouy:
Hey all!

Any idea when this PR could be merged ? It would really help us to clean how state is handled in fundraising.

Because of the tap lowering the pool reserve every block we basically need to re-call `balance` on the reserve every 15 seconds and then update a bunch of state entry based on that ... It would be way cleaner if we could trigger the state update to go through the background script and app reducer ...

packages/aragon-api/src/index.js
```diff
@@ -547,6 +548,31 @@ export class AppProxy {
...
+    return this.rpc.send(
+      'trigger',
+      [eventName, returnValues]
+    )
+  }
+
+  /**
+   * subscribe to an observable that emits events created by the frontend event triggers
+   */
+  frontendTriggers () {
```
sohkai: Perhaps we can rename this to just `triggers`, or perhaps `observeTriggers`.

I'm also not sure about thinking about this as a solely `frontend -> background script` type of action; my original thought was to let this be generic and bi-directional. However I hadn't considered how the sending side would ignore its own messages, which might be tricky (and perhaps not worthwhile) to solve.

packages/aragon-wrapper/src/index.js
```diff
@@ -231,6 +231,7 @@ export default class Aragon {
     this.initAppIdentifiers()
     this.initNetwork()
     this.initNotifications()
+    this.trigger = new Subject()
```
sohkai: Rather than having a global "trigger" message bus, I've been thinking more and more about having application-specific contexts to manage this better (ie. anything that's not really relevant for a client to subscribe to).

A few things fit:

- App triggers
- App state (the localstorage / indexeddb caching will be turned into a side effect of the state soon)
- Perhaps app paths

And maybe more.

sohkai: An initial version of this context has been implemented in https://github.com/aragon/aragon.js/pull/352 ([implementation](https://github.com/aragon/aragon.js/blob/7e8f263d8610fdb038427535b8eb9d3f1e8747ae/packages/aragon-wrapper/src/apps/index.js)).

topocount:
> I know Fundraising needs this feature a bit more urgently than I'd like (the issue's been open for more than a year, so there's been ample time to have signalled this requirement wink), so we may have to release early and make some choices that will let us work towards a final implementation without backwards compatibility issues.

While it may be true that this issue has been open for over a year, a workaround was available (and utilized by Rewards and Projects) until #294 was implemented, so this PR only became urgently needed a couple months ago.

>Rather than having a global "trigger" message bus, I've been thinking more and more about having application-specific contexts to manage this better (ie. anything that's not really relevant for a client to subscribe to).

We can discuss this more on the all-devs call tomorrow, but I was initially planning to implement this entirely within the aragonAPI and bypass the wrapper. Happy to go back to this implementation if you think it makes more sense.

sohkai:
> We can discuss this more on the all-devs call tomorrow, but I was initially planning to implement this entirely within the aragonAPI and bypass the wrapper.

There is not really a way to do this; aragonAPI without a wrapper in-between is basically two dead ends. The RPC flow is designed to always go through the wrapper module (unless we start associating the running instances of an app together, but then a whole other set of APIs need to be built).

docs/API.md
```diff
@@ -162,16 +164,19 @@ A simple representation of the store's lifecycle:
...
-1. Fetch past events from the contract and any given `options.externals` contracts, reducing new state from found events. Note that new events from these contracts are **not** fetched until all past events have been found.
-1. Cache the state as "committed" state, as before
+1. Fetch past events from the application contract and any given `options.externals` contracts, reducing new state from found events. Note that new events from both the application contract and external contracts are **not** fetched until all past events have been found.
+1. Cache the state at the end of the initial sync as "committed" state, as before
 1. Subscribe to new events from the contract and any given `options.externals` contracts, reducing new state based on the incoming events.
 
+If the application emits its own `trigger`s (see [`emitTrigger`](#emittrigger)), the triggers will appear in the reducer only once the syncing phase for past events ends. Any triggers emitted before this will be ignored. Triggers are implicitly transformed into the web3 event schema of `event` (string) and `returnValues` (object).
```
sohkai: > The triggers will appear in the reducer only once the syncing phase for past events ends. Any triggers emitted before this will be ignored

This is a pretty big limitation in the current implementation, given that the initial sync can be quite lengthy. One way to get around this is to only emit triggers from the frontend once you've detected the sync period is over (`SYNC_STATUS_SYNCED`), but this workaround will not work in all cases (e.g. when trying to access an external API before the initial sync is over) and is somewhat silly.

I have a change to `store()` that will fix this, so triggers will be emitted during the past events' sync phase as well, but this is a more sensitive change and I'd like to test it before pushing.

Quazia: I think this is fine for now, since in most of our use cases you'll want to allow the past events to be synced before interacting with the application anyways.

sohkai: Confirmed a96b0f4 works for propagating triggers during the past events phase.

Due to the way the events are streamed, all past events from a contract are handled "in order" and don't give up control to a trigger that is queued. This is most noticeable if you have a trigger emitting on a timed interval, where on the past events being fetched, the timed trigger will act as if it's been paused (in reality it's just getting queued and will all be flooded into the reducer when the past events have finished reducing).

sohkai:
Found out that the previous `store()` implementation was a bit naive with how it was treating `trigger()`s (there would be a gap in listening to the `trigger()`s in-between past events getting fetched and finishing reducing state from past events; sometimes this can take a noticeable while).

Added a8cb3bc to refactor how `store()` works. This greatly simplifies the logic, and makes it easy to handle `trigger()`s consistently.

PR #364, opened 2019-08-28
API: update behaviour of store() to be more consistent

sohkai:
**Changes**

Updates the `api.store()` behaviour to be a bit more consistent / less surprising.

Changes:

- ⚠️ If `options.init()` is available, use only its returned state to start the reducer (rather than merging its result with the cached state, which may be surprising)
- Make sure `null` is used as the initial state for the `reducer` as well as `options.init`
- Make sure `options.init()` cannot accidentally manipulate the object held in the cache observable
- Improvements to the debug logging

- [x] I have updated the associated documentation with my changes (see #359).


2color:
LGTM.

PR #380, opened 2019-09-03
Api: add iconSrc of installed apps

sohkai:
Adds an `iconSrc` property to any returned app objects from `api.installedApps()` and `api.currentApp()` (if the icons are available).

If there are multiple icons declared in an app's `manifest.json`, this uses the first one for simplicity.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -253,11 +253,12 @@ Returns **[Observable](https://rxjs-dev.firebaseapp.com/api/index/class/Observab
 - `appAddress`: this app's contract address
 - `appId`: this app's appId
 - `appImplementationAddress`: this app's implementation contract address, if any (only available if this app is a proxied AragonApp)
-- `iconSrc`: a link to this app's icon asset, if any
 - `identifier`: this app's self-declared identifier, if any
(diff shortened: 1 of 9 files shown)


bpierre:
I wonder if we shouldn’t also (or only?) expose a `.getIconSrc(size)` method, doing what [getAppIconBySize()](https://github.com/aragon/aragon/blob/658c5282ed6e1644245e0ee78a73727f4f7b2a01/src/utils.js#L11-L38) does in aragon/aragon?

sohkai:
Mmm so you'd give back both the `content` and the `icons` and let the users figure out which one they'd like to use?

bpierre:
I think we can provide these, and let users only set the display size they want − or no parameters at all to get the default size.

```jsx
<img width="40" height="40" src={app.iconSrc(40)} />
```

sohkai:
@bpierre I've adjusted the icon API to be based on the size-based api in 4384998.

For now I think it's OK to copy over the code; it's fairly small and generic code for handling [web manifest icons](https://developer.mozilla.org/en-US/docs/Web/Manifest/icons).

PR #390, opened 2019-09-26
Wrapper: clarify calculateTransactionPath() error when destination is not installed

macor161:
**Changes**

Small change to throw a more specific error message when `app` is not found. 

The previous error was thrown in `createDirectTransactionForApp` and was a little bit misleading: `Could not create transaction due to missing app artifact`. People often thought it was a problem related to IPFS.

- (N/A) I have updated the associated documentation with my changes


macor161:
@sohkai Thanks for the adjustment! Yes exactly, and also when trying to execute a function on an app that doesn't have any permissions.

PR #391, opened 2019-09-30
Feat/address book identity provider

topocount:
**Changes**

This Identity Provider looks through all apps for address book instances and then returns the first result for the resolver. Search functionality has also been implemented in order to take advantage of autocomplete functionality


lkngtn:
In the process of migrating 1Hive to mainnet and because this hasn't been merged yet we are losing address book resolution for the allocation process. We should be able to make things work by having everyone import custom labels, but it also means that each time we add someone everyone needs to take the action to update their labels which is not ideal... 

It would be amazing to get this in and if there is anything we can do to help please let us know!

PR #392, opened 2019-09-30
Wrapper: Add error log when radspec evaluation fails

### LINKED ISSUE #101: Gracefully handle errors in Radspec
1. Ignore error and provide a default message
2. Log error to console
3. ???
4. Profit

gh1dra:
Fixes #101.

We could also take this opportunity to add tests to `describeTransaction` but I'm not sure where to put them so that they are aligned with the pattern. Also creating stubs for radspec could be additional work outside the scope of this small PR.

### DIFF
--- a/packages/aragon-wrapper/src/rpc/handlers/describe-transaction.js
+++ b/packages/aragon-wrapper/src/rpc/handlers/describe-transaction.js
@@ -11,7 +11,8 @@ export default async function (request, proxy, wrapper) {
 
   let description
   try {
-    description = await tryEvaluatingRadspec(transaction, wrapper)
+    const decoratedTransaction = await tryEvaluatingRadspec(transaction, wrapper)
+    description = decoratedTransaction.description
   } catch (_) {}
 
   if (description) {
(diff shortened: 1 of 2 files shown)


packages/aragon-wrapper/src/rpc/handlers/describe-transaction.js
```diff
@@ -22,6 +22,10 @@ export default async function (request, proxy, wrapper) {
         description: processed.description
       }
     } catch (_) {}
+  } else {
+    return {
+      ...transaction
+    }
```
gh1dra: Was trying to encapsulate the logic here: https://github.com/aragon/aragon.js/issues/101#issuecomment-418801302

Though this may not be needed, as if the `evaluatedNotice` is undefined, the `tryEvaluatingRadspec` method returns the fields inside the transaction as well: https://github.com/aragon/aragon.js/blob/1684b7696d48ed3e248ba1dc426e343dff5ed850/packages/aragon-wrapper/src/radspec/index.js#L52

sohkai: Yes, this part isn't needed because we're only interested in returning the `description` or `annotatedDescription` in this API and not the entire transaction (it's already given as input).

PR #408, opened 2019-11-27
api: add GUI style

bpierre:
Allows to share an appearance and a theme with the apps running.

`appearance` is set to either `light` or `dark`. Other values could be passed in the future (e.g. `black` for OLED screens). It is always present and apps should respect it and display a theme that correspond to it.

`theme` contains the entire theme that should be rendered by the app. It is optional and apps should respect it when present. If not possible, apps should respect the `appearance`.

Theme examples:

https://github.com/aragon/aragon-ui/blob/master/src/theme/theme-light.js
https://github.com/aragon/aragon-ui/blob/master/src/theme/theme-dark.js


bpierre:
:checkered_flag:

PR #415, opened 2019-12-15
Use new pacakges @aragon/abis to fetch aragonOS abis

0xGabi:
**Changes**

Remove `@aragon/os` as dependency and start using `@aragon/abis` to fetch ABIs needed.


sohkai:
Going to leave this closed; I think `@aragon/os` should be really fast to install once we've removed the built-in truffle dependents.

PR #424, opened 2020-02-02
Wrapper: support ACL intents for api.external handles

### LINKED ISSUE #417: Wrapper: Enable ACL intents in apps
Calling some ACL functions from inside an app don't work, for example:
```js
const aclHandler = api.external(aclAddr, aclAbi)
aclHandler.revokePermission(entity, app, role).toPromise()
```
Doesn't find a transaction path because `revokePermission` have not an `auth()` modifier (it has an `onlyPermissionManager()` modifier instead).

I propose to change the behavior of the [`intent` handler](https://github.com/aragon/aragon.js/blob/b94f002b28ab9f021e1bd3db690e049a3841451c/packages/aragon-wrapper/src/rpc/handlers/intent.js) so it does a [`performACLIntent()`](https://github.com/aragon/aragon.js/blob/b94f002b28ab9f021e1bd3db690e049a3841451c/packages/aragon-wrapper/src/index.js#L1230) if needed.

sembrestels:
Fixes #417.

**Changes**

Support ACL transaction path discovery when performing actions in the ACL obtained from `api.external()` such as:
```
const aclHandler = api.external(aclAddr, aclAbi)
aclHandler.revokePermission(entity, app, role).toPromise()
```

It doesn't change the API

### DIFF
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -1329,6 +1329,10 @@ export default class Aragon {
   async getExternalTransactionPath (destination, methodJsonDescription, params) {
     let path
 
+    if (addressesEqual(destination, this.aclProxy.address)) {
+      return this.getACLTransactionPath(methodJsonDescription.name, params) || []
+    }
+
     const installedApp = await this.getApp(destination)
     if (installedApp) {
       // Destination is an installed app; need to go through normal transaction pathing
--- a/packages/aragon-wrapper/src/index.js
+++ b/packages/aragon-wrapper/src/index.js
@@ -1327,28 +1327,30 @@ export default class Aragon {
    *   single transaction.
    */
   async getExternalTransactionPath (destination, methodJsonDescription, params) {
-    let path
-
     if (addressesEqual(destination, this.aclProxy.address)) {
-      return this.getACLTransactionPath(methodJsonDescription.name, params) || []
+      try {
+        return this.getACLTransactionPath(methodJsonDescription.name, params)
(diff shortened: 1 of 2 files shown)


packages/aragon-wrapper/src/rpc/handlers/external.js
```diff
@@ -27,11 +28,13 @@ export async function intent (request, proxy, wrapper) {
     ...params
   ] = request.params
 
-  const transactionPath = await wrapper.getExternalTransactionPath(
-    address,
-    methodJsonDescription,
-    params
-  )
+  const transactionPath = addressesEqual(address, wrapper.aclProxy.address)
+    ? await wrapper.getACLTransactionPath(methodJsonDescription.name, params)
```
sohkai: Actually, I'm wondering if this would make sense to move this logic to `getExternalTransactionPath()`, since the ACL is always an "external" destination to apps running against the RPC.

sembrestels: Yes, it has total sense.

sembrestels:
Code moved to `getExternalTransactionPath()` and test added.

sohkai:
@sembrestels Sorry this took a bit to land into the client; we were busy with a few other efforts surrounding the Court.

This is now available as part of 0.8.8, which you can grab and run now, or otherwise wait to use it from a new version of the CLI (v7.1.1) that will be published in the next day or so.

sembrestels:
Thanks @sohkai. I've just tested it and everything works as expected!

Issue #423, opened 2020-01-31
Wrapper: merge AppProxy's ABI into app's ABI if we know app is a proxy


sohkai:
See original comments in https://github.com/aragon/aragon-cli/issues/659.

Rather than doing this at "build" time, it makes more sense to do at run time, when we can check whether or not the app is actually a proxy (checking `implementation()`).

We may also want to add an `isProxy` flag to the app object to help others distinguish if it is a proxy.

PR #427, opened 2020-02-18
Wrapper: merge app ABI with AppProxy ABI

alexmedkex:
**Changes**

- Include contract functions from proxy contract ABI in the app contract ABI.
- Function signature collision check when instantiating a proxy app.

This is addressing https://github.com/aragon/aragon.js/issues/423

Note: currently using console.log() to warn user of collisions.

Regarding checking if the app is using a proxy; the `runApp()` method  seems to assume that it is, since it uses a `proxyAddress`. Should this method be changed to not have this assumption?

### DIFF
--- a/packages/aragon-wrapper/src/utils/index.js
+++ b/packages/aragon-wrapper/src/utils/index.js
@@ -64,24 +64,24 @@ export function makeProxy (address, interfaceName, web3, options) {
 const appProxyAbi = getAbi('aragon/AppProxy').filter(({ type }) => type === 'event' || type === 'function')
 export function makeProxyFromAppABI (address, appAbi, web3, options) {
   const collisions = findFunctionSignatureCollisions(appAbi, appProxyAbi)
-  if(collisions.length > 0) {
+  if (collisions.length > 0) {
     console.log(
       `WARNING: Collisions detected between the proxy and app contract ABI's.
        This is a potential security risk.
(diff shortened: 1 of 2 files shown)


sohkai:
> Function signature collision check when instantiating a proxy app.

This particular step should also be done at build step.

I agree there's not much we can really do about at run time, outside of setting a flag somewhere in the app object to alert users of a potential security issue.

alexmedkex:
> > Function signature collision check when instantiating a proxy app.
> 
> This particular step should also be done at build step.
> 
> I agree there's not much we can really do about at run time, outside of setting a flag somewhere in the app object to alert users of a potential security issue.

It looks to me that the app artifact generation is done outside of aragon.js, or what am I missing?

For example here https://github.com/aragon/buidler-aragon/blob/master/src/utils/generateArtifacts.ts

alexmedkex:
@sohkai 

I have added the build step part here:
https://github.com/aragon/aragon-cli/pull/1537

Should this be added into `aragon-buidler` package as well?

sohkai:
@alexmedkex Sorry for taking so long on coming back to this! Will review soon!

> Should this be added into aragon-buidler package as well?

Yes, we should also add the signature collision check to the buidler plugin too!

alexmedkex:
No worries, although I think this is now outdated.

I added the collision check to the aragon-cli toolkit https://github.com/aragon/aragon-cli/pull/1537, so that can probably be re-used here..

Issue #362, opened 2019-08-27
API: allow filtering past events by indexed parameters


sohkai:
There may be some cases, such as searching for token transfers to a specific address, where it would be useful to optimize the past logs fetching as much as possible.

We should allow developers using the `external.pastEvents()` API to explicitly declare more topics to narrow down a search. On token transfers, for example, a developer should be able to specify a filter for just the from address or the to address.

vporton:
This is not an `enhancement` but a bug: [Docs](https://hack.aragon.org/docs/api-js-ref-api) says that `pastEvents` has the same interface as `web3.eth.Contract.getPastEvents()` which does have `filter` parameter.

vporton:
It appears that filtering by `event` also does not work!

```
this.courtNamesContractHandle.pastEvents({event: 'SetCourtName', fromBlock: 0, filter: {ourCourtId: 1}})
```

vporton:
I want to fix this bug.

But before please explain in `packages/aragon-wrapper/src/rpc/handlers/external.js`:

```
export function pastEvents (request, proxy, wrapper) {
...
{
```

Why does this function appear to be not a member of a contract object but a standalone function? Why does it not use `this`? How is it called when one calls `contract.pastEvents`? What do the parameters of this function mean?

vporton:
It seems I fixed the bug: https://github.com/aragon/aragon.js/pull/428

(but need to add a unit test, what I am myself unsure how exactly to do)

sohkai:
> It appears that filtering by event also does not work!

Yes, we differ from the web3.js API here, where we do not provide a parameter to specify the event name. Because of this, and to not create confusion with another API, fixing this will be a breaking change and require a v3 of the API (not currently planned), and will be tracked in https://github.com/aragon/aragon.js/issues/363.

> We should allow developers using the external.pastEvents() API to explicitly declare more topics to narrow down a search

On my original comment, I believe the `filter` option should work right now, but it is not ideal because both `topics` and `filter` usually apply to only one or a subset of events.

In any case, we'll track this story in https://github.com/aragon/aragon.js/issues/363.

PR #428, opened 2020-02-21
https://github.com/aragon/aragon.js/issues/362 - fix (untested)

vporton:
**Changes**

It seems I fixed issue #362.

However, I don't know how to properly test it. Please test.


sohkai:
Please see response in https://github.com/aragon/aragon.js/issues/362#issuecomment-605885794.

I would not like to add `event` as an event option, because ideally it would be a separate parameter.

PR #434, opened 2020-05-04
Add new MANAGE_PROFILE_ROLE for DAO Profiles

Evalir:
**Changes**
Modifies the Kernel.json artifact, adding a new role with the name `Manage profile` and id `MANAGE_PROFILE_ROLE`.

### DIFF
--- a/.size-limit.js
+++ b/.size-limit.js
@@ -12,7 +12,7 @@ module.exports = [
   {
     name: '@aragon/wrapper',
     path: "packages/aragon-wrapper/dist/index.js",
-    limit: "525 KB"
+    limit: "540 KB"
   },
   {
     name: '@aragon/api-react',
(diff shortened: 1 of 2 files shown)


Evalir:
Yeah, for now, let's add (re-add? I remember seeing a (very funny) HACKS.md file before 😃) a HACKS.md file so people are aware.

sohkai:
:+1: We can do so in a separate PR. That used to be in the aragon/aragon repo and was removed because we had no leftover hacks to document, but we can introduce it here to keep track of things like this.

cygnusv:
Is there any documentation anywhere of what this role does?

sohkai:
@cygnusv Not really; this was a "fake" role that we used for setting up DAO profiles on apiary.

PR #436, opened 2020-06-10
Wrapper: update post-processed radspec to use fancy quotes

sohkai:
Updates the descriptions with these nicer quotes than `'` :).


0xGabi:
Taking this 😏

Issue #412, opened 2019-12-09
Wrapper: consolidate utilities to find a method's JSON description and function artifact


sohkai:
We have a number of duplicated places where we're doing a `.find()` based on a given method signature (either simple, e.g. `getVote` or full, e.g. `getVote(uint256)`) to find a method's JSON description (from the app's ABI) or the method's function artifact (from the app's artifact.json; usually for finding roles or radspec).

This happens at least when calculating transaction paths, both normally as well as for ACL intents.

PR #443, opened 2020-07-17
wrapper: consolidate and deduplicate ABI and method searching

sohkai:
Fixes https://github.com/aragon/aragon.js/issues/412.

Deduplicates ABI or artifact searching logic where possible. The consolidation also allows us to make sure we handle some edge cases correct, e.g. fallback functions or incomplete method signatures (where we use method names rather than full signatures).

It's still not the nicest as we don't clearly delineate when an ABI fragment or a raw method signature should be used due to everything being a public interface, so we end up doing the method search multiple times.

### DIFF
--- a/packages/aragon-wrapper/src/utils/abi.js
+++ b/packages/aragon-wrapper/src/utils/abi.js
@@ -1,5 +1,6 @@
 export function findMethodAbiFragment (abi, methodSignature) {
   if (methodSignature === 'fallback') {
+    // Note that fallback functions in the ABI do not contain a `name` or `inputs` key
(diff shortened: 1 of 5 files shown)


packages/aragon-wrapper/src/utils/abi.js
```diff
@@ -0,0 +1,26 @@
+export function findMethodAbiFragment (abi, methodSignature) {
+  if (methodSignature === 'fallback') {
+    return abi.find(method => method.type === 'fallback')
```
sohkai, line 3: Note for future: fallback functions do not contain `name` or `inputs` fields in the ABI.

PR #450, opened 2021-02-17
Static Cache Tracker utility

### LINKED ISSUE #400: Wrapper: add utility to clear localstorage and IndexedDB caches
There are various hiccups with IndexedDB:

- [Firefox does not support `indexedDB.databases` yet](https://bugzilla.mozilla.org/show_bug.cgi?id=934640)
- aragon.js may switch to localStorage in some environments if IndexedDB is not secure

This makes it harder for users of the wrapper API to know how to reset their state, outside of resetting all of localStorage and IndexedDB.

Instead, the wrapper should expose a simple static utility that cleans up all of its state. To do this, the wrapper should keep, in localStorage, the set of cache keys it uses.

namesty:
This PR aims to close #400.

Currently, localForage is being used for caching in @aragon/wrapper. And while keys for each DB is being saved so that they can be cleared later, the actual DB instances persisted in the browser's IndexedDB/Storage. Also, since trackedKeys are being saved in memory and not being persisted, a refresh makes the app lose track of its previous keys.

I created a static utility that creates a localForage instance that saves each DB instance created and persists their names, so that they can be later retrieved and erased without needing to wipe the browser's complete IndexedDB. I am exposing this utility for the client to use.

This change is non-disruptive in regards to the current interfaces.

I just need to make the tests for it

### DIFF
--- a/packages/aragon-wrapper/src/cache/index.js
+++ b/packages/aragon-wrapper/src/cache/index.js
@@ -19,8 +19,8 @@ import * as configurationKeys from '../configuration/keys'
 export default class Cache {
   static cacheTracker
 
-  static async getCacheTracker() {
-    if(!Cache.cacheTracker) {
+  static async getCacheTracker () {
+    if (!Cache.cacheTracker) {
       await Cache.initCacheTracker()
     }
 
@@ -29,7 +29,7 @@ export default class Cache {
 
   /**
     * Initializes the singleton cache tracker
-    * 
+    *
     * @returns {Promise<void>}
     */
   static initCacheTracker = async () => {
@@ -39,7 +39,7 @@ export default class Cache {
 
   /**
     * Returns currently saved caches's keys, without duplicates
-    * 
+    *
     * @returns {Promise<string[]>}
     */
   static getSavedCaches = async () => {
@@ -52,13 +52,13 @@ export default class Cache {
     * Save new cache's key to track it and delete it later
     * @param {string} prefix
     *       String prefix to use for the cache
-    * 
+    *
     * @returns {Promise<void>}
     */
   static trackNewCache = async (prefix) => {
     const tracker = await Cache.getCacheTracker()
     const savedCaches = await this.getSavedCaches()
-    if(savedCaches[0]) {
+    if (savedCaches[0]) {
       await tracker.set(configurationKeys.CACHE_TRACKER_CACHES_KEY, [...savedCaches, prefix])
     } else {
       await tracker.set(configurationKeys.CACHE_TRACKER_CACHES_KEY, [prefix])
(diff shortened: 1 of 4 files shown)


packages/aragon-wrapper/src/cache/index.js
```diff
@@ -17,6 +17,52 @@ import * as configurationKeys from '../configuration/keys'
...
+
+    const savedCaches = await this.getSavedCaches()
+    if(savedCaches[0]) {
+      await Cache.cacheTracker.set(configurationKeys.CACHE_TRACKER_CACHES_KEY, [...savedCaches, prefix])
+    } else {
+      await Cache.cacheTracker.set(configurationKeys.CACHE_TRACKER_CACHES_KEY, [prefix])
+    }
+  }
+
+  static clearAllCaches = async () => {
```
nivida, line 47: I would use for all those methods the "normal" syntax e.g.:
``static async clearAllCaches() {...}

packages/aragon-wrapper/src/cache/index.js
```diff
@@ -17,6 +17,52 @@ import * as configurationKeys from '../configuration/keys'
  *        Require the cache to downgrade to localstorage even if IndexedDB is available
  */
 export default class Cache {
+  static cacheTracker
+
+  static initCacheTracker = async () => {
+    if(!Cache.cacheTracker) {
```
nivida, line 23: This singleton implementation could get improved with a ``Cache.getTracker()``. Such a method would remove some code duplications here in this class.

packages/aragon-wrapper/src/cache/index.js
```diff
@@ -17,6 +17,52 @@ import * as configurationKeys from '../configuration/keys'
  *        Require the cache to downgrade to localstorage even if IndexedDB is available
  */
 export default class Cache {
+  static cacheTracker
```
nivida, line 20: Is it possible for you to add JSDoc's anywhere here.. an example of how I was doing them: 
https://github.com/aragon/govern/blob/feature/govern-tx/packages/govern-tx/src/wallet/Wallet.ts#L36-L48

packages/aragon-wrapper/src/configuration/keys.js
```diff
@@ -1,2 +1,4 @@
 export const FORCE_LOCAL_STORAGE = Symbol('FORCE_LOCAL_STORAGE')
 export const SUBSCRIPTION_EVENT_DELAY = Symbol('SUBSCRIPTION_EVENT_DELAY')
+export const CACHE_TRACKER_PREFIX = "cacheTracker"
+export const CACHE_TRACKER_CACHES_KEY = "caches"
```
nivida, line 4: EOF

nivida (changes requested):

PR #483, opened 2021-08-24
DAO-212: fix private network type issue with polygon mumbai

yuetloo:
@novaknole This change pulls in use-wallet's KNOWN_CHAINS to map mumbai's network type, noticed a lot of ethersproject dependencies pulled in as a result... please review if it's ok.


yuetloo:
Hmm... the size of wrapper almost doubled after pulling use-wallet causing the CI main to fail.. the other CI may be related to using branch instead of npm package... will test it out later..

PR #488, opened 2021-09-29
update: added support for Harmony One Blockchain

victaphu:
**Changes**
Changed the way events are retrieved because Harmony One wss end-points restrict the amount of blocks scanned to 1024. Added a parameter called PAST_EVENTS_BATCH_SIZE that indicates the batch size for block scans when using getPastEvents.


victaphu:
I'll redo this pull, the CLA is not signed and the user zain is a misconfigured user on my system

PR #489, opened 2021-09-29
update: added support for block size limit on getPastEvents

victaphu:
Harmony One limits the number of blocks returned by getPastEvents to only 1024. Beyond that and the API throws an exception.

**Changes**

### DIFF
--- a/packages/aragon-wrapper/src/core/proxy/index.js
+++ b/packages/aragon-wrapper/src/core/proxy/index.js
@@ -29,9 +29,27 @@ export default class ContractProxy {
 
     // The `from`s only unpack the returned Promises (and not the array inside them!)
     if (eventNames.length === 1) {
+      if (!process.env.PAST_EVENTS_BATCH_SIZE) {
+        return from(this.contract.getPastEvents(eventNames[0], options))
+      }
       // Get a specific event or all events unfiltered
       return from(
-        this.contract.getPastEvents(eventNames[0], options)
+        new Promise(async resolve => {
+          // console.log("Resolving ", resolve);
+          const ranges = [];
+
+          for (let i = +options.fromBlock; i < +options.toBlock; i += process.env.PAST_EVENTS_BATCH_SIZE) {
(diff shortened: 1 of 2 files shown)


packages/aragon-wrapper/src/core/proxy/index.js
```diff
@@ -29,9 +29,27 @@ export default class ContractProxy {
...
+        return from(this.contract.getPastEvents(eventNames[0], options))
+      }
       // Get a specific event or all events unfiltered
       return from(
-        this.contract.getPastEvents(eventNames[0], options)
+        new Promise(async resolve => {
+          // console.log("Resolving ", resolve);
+          const ranges = [];
+
+          for (let i = +options.fromBlock; i < +options.toBlock; i += process.env.PAST_EVENTS_BATCH_SIZE) {
```
WissenIstNacht, line 41: Hey, could you please provide some context as to why this is necessary?

victaphu: Hi @WissenIstNacht, harmony one blockchain web socket restricts the number of blocks returned for any getPastEvents call to 1024 blocks. Beyond that it will throw an exception (get_Logs limited to 1024 blocks). If the above code is not in place, then we get exceptions whenever Aragon tries to get past events.

I spoke to the Harmony engineers who confirmed this as a configuration on their web socket endpoint.

nivida: Thanks a lot for your PR and the explanation here. I have a few suggestions to the submitted code.

1. The new Promise() isn't required cause of the async keyword. You can define the inline function as async and just return the values. The return is the same as resolve()
2. Instead of having two loops can you do this within one loop
3. The options object doesn't have to be created N times (will use some memory) 
4. The out commented console.log can get removed 
5. This logic got duplicated in external.js so it would be great to move this out as a function that can get called on both places.
6. The error handling could get improved to provide for example more information on which step it was failing.

Feel free to challenge my requested changes. Happy to get corrected accordingly in case I missed something ;-)

nivida (changes requested):

